PR #17, opened 2026-05-18
Read research vault path from OBSIDIAN_VAULT_PATH

jameswolensky:
The research toolkit hardcoded VAULT_PATH to a personal path in scripts/research/lib/config.py, so every other user's /youtube, /research, /research-deep and /notebooklm notes saved into a nonexistent vault and the printed obsidian:// links failed to resolve.

This reads the path from OBSIDIAN_VAULT_PATH — already the project-wide convention (hooks, setup.sh, quick-install.sh, SKILL.md) — via the existing get_required helper, so a missing value fails fast with the standard message instead of silently saving to the wrong place. .env.example documents it as required, and setup.sh now also writes it into the research .env, so it resolves on standalone CLI runs and not only when it is already present in the environment.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -2,6 +2,10 @@
 # Copy this file to ~/.config/obsidian-second-brain/.env and fill in the values.
 # That path is Mac-local — never committed, never synced. Set permissions 600.
 
+# ─── Vault location (required) ──────────────────────────────
+# Absolute path to your Obsidian vault root. Research notes save here.
+OBSIDIAN_VAULT_PATH=
+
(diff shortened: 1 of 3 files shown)


eugeniughelbur:
thanks @jameswolensky, clean fix. that hardcoded path was silently breaking the research toolkit for everyone but me, the obsidian:// links wouldn't resolve either. good call routing it through `get_required` so it fails fast, and nice catch wiring `.env.example` + `setup.sh` in the same PR so standalone CLI runs work too. merged.

Issue #1, opened 2026-04-19
PostCompact background agent never runs — stdin schema mismatch causes silent early exit


sad85520:
## TL;DR

On Claude Code `2.1.112`, `hooks/obsidian-bg-agent.sh` exits at the
`[[ -z "$SUMMARY" ]] && exit 0` guard on every `PostCompact` event without
spawning the headless `claude -p` subprocess, so no vault propagation happens.
The failure is invisible because Claude Code reports `completed successfully`
for any exit 0 — indistinguishable from a successful no-op.

## Root Cause

The script expects the PostCompact event to inline a compaction summary
in stdin:

```bash
SUMMARY=$(printf '%s' "$INPUT" | jq -r '.summary // .content // ""' 2>/dev/null || true)
[[ -z "$SUMMARY" ]] && exit 0
```

The actual `PostCompact` stdin includes `transcript_path` (verified by the
patched version working end-to-end) but does not include `.summary` or
`.content` (confirmed by the always-empty `SUMMARY` and silent exit — see
Claude Code hook docs for the full schema). So `SUMMARY` is always empty
→ `exit 0` → `claude -p` is never invoked → `/tmp/obsidian-bg-agent.log`
is never created.

## Evidence

Reproduced on Claude Code `2.1.112`:

1. Triggered `/compact` six times in a single session over several hours.
2. Each time Claude Code displayed
   `PostCompact [...obsidian-bg-agent.sh] completed successfully`.
3. `/tmp/obsidian-bg-agent.log` did not exist afterwards.
4. No writes in the vault corresponding to the sessions' content.

Piping a fabricated payload manually confirmed the script's downstream
reaches `claude -p` once `SUMMARY` is non-empty:

```bash
echo '{"summary":"test"}' | OBSIDIAN_VAULT_PATH=/vault bash hooks/obsidian-bg-agent.sh
# ↳ spawns headless claude, writes /tmp/obsidian-bg-agent.log
```

## Proposed Fix

The compaction summary is stored in the transcript JSONL as entries with
`isCompactSummary: true`. Read the most recent one from `transcript_path`:

```diff
- # Parse compaction summary from stdin JSON
+ # PostCompact stdin includes `transcript_path`; the compaction summary itself
+ # is written into the transcript JSONL as entries with `isCompactSummary: true`.
+ # We read the most recent one here.
  INPUT=$(cat)
[...]

PR #19, opened 2026-05-20
Add /projects, /obsidian-setup, Bases, SessionStart hook, per-day logs

MPZ-00:
## What this PR does

Adds a developer-focused layer on top of the existing 33 commands: a git-aware project dashboard (`/projects`), a runtime per-user command opt-out system (`/obsidian-setup`), folder
Bases for live vault views, a SessionStart hook that injects `_CLAUDE.md` once per session instead of re-reading it on every command, and a per-day log structure under `Logs/` with
supporting scripts.

## Type of change

- [x] 📜 New slash command
- [x] 📚 Documentation update

## Related issues

n/a

## How I tested it

- `ls commands/` confirms 35 files (33 upstream + projects + obsidian-setup)
- `bash scripts/build.sh --platform claude-code` completes cleanly; `dist/claude-code/commands/` has 35 commands
- `python scripts/vault_stats.py --help` and `python scripts/migrate_log.py --help` run without import errors
- `/obsidian-setup` run in test vault — confirmed `disabled_commands:` written to `_CLAUDE.md` and respected on next command invocation
- `/projects` run with a populated `projects:` block — confirmed parallel subagents fired, `Projects/Dashboard.md` rewritten, dashboard printed to conversation

- [x] If I added or changed a command that writes to the vault, the produced note has a `## For future Claude` preamble
- [x] Wikilinks (`[[...]]`) used for every person, project, idea, decision referenced
- [x] External claims have recency markers (`(as of YYYY-MM, source.com)`)
- [x] Sources preserved verbatim (URLs inline, not paraphrased)

## Checklist

- [x] If I added a new command, I added an entry in `SKILL.md` and the README's command table

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ __pycache__/
 # Local config (never commit)
 .env
 *.local
+.claude/
 
 # macOS
 .DS_Store
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 
 ## [Unreleased]
 
+### Added
+
+- **SessionStart hook (`hooks/load_vault_context.py`):** injects `_CLAUDE.md` into context once per session when the session starts inside the vault. Eliminates the per-command re-read of `_CLAUDE.md` that burned tokens on every invocation. Wired automatically by `scripts/setup.sh`.
+- **`scripts/setup.sh` updated:** wires the new SessionStart hook (`hooks/load_vault_context.py`) in addition to the existing PostCompact background agent.
+
 ## [0.8.0] — 2026-05-15
 
 ### Added
--- a/SKILL.md
+++ b/SKILL.md
@@ -31,6 +31,10 @@ description: >
 
 Try these methods in order. Use the first one available:
 
+**Method 0 — SessionStart hook (if configured):**
+If `hooks/load_vault_context.py` is wired as a SessionStart hook in `~/.claude/settings.json`, `_CLAUDE.md` is injected into context automatically at session start. Skip step 1 below.
+To wire it: `bash scripts/setup.sh "/path/to/vault"` or run `/obsidian-setup`.
+
 **Method A — MCP server (`mcp-obsidian`):**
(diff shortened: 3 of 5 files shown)


eugeniughelbur:
thanks @MPZ-00, real thinking here. the SessionStart hook alone is the single best token optimization the skill could ship right now (someone just hit me on X yesterday about the 14% startup cost). per-day logs and the `disabled_commands` opt-out are also solid ideas.

but this is too much to land as one PR. a few specific issues worth flagging:

- `commands/obsidian-init.md` references `~/.claude/skills/obsidian-ai-brain/` in two places — the skill is `obsidian-second-brain`, looks like a local-fork name leaked through.
- the hook builds the context header with hardcoded `\\` separators (e.g. `{vault}\\_CLAUDE.md`), which renders as literal backslashes on macOS / linux. needs to use `os.sep` or just forward slashes.
- the new vault folder structure (`Projects/`, `People/`, `Knowledge/` at vault root, plus 8 capitalized Bases targets) differs from the convention this skill ships and that existing users run with (`wiki/projects/`, `wiki/entities/`, `wiki/concepts/`). would break existing vaults without an explicit migration path.
- `scripts/_strip_command_boilerplate.py` strips "Read `_CLAUDE.md` first" and the AI-first footer from every `obsidian-*.md` command file. that breaks the codex / gemini / opencode adapters (no SessionStart hook on those platforms) and explicitly violates the repo `CLAUDE.md` rule "do not strip frontmatter or `## For future Claude` preambles from existing commands."

can you split this into separate PRs:

1. SessionStart hook + `setup.sh` wiring (with the backslash fix). this one i'll merge fast — it's the highest-value piece.
2. per-day logs + `migrate_log.py` + `vault_stats.py`.
3. `/obsidian-setup` + `disabled_commands:` config support.
4. `/projects` + folder Bases (plus a separate discussion on whether the skill migrates to the new vault structure or whether these stay structure-agnostic).

happy to land #1 quickly. the others i'd want to evaluate independently and probably discuss the vault-structure question before merging anything that assumes the new layout.

MPZ-00 (commented):
Alright, thx, I'll update it to be a lot less :)

Issue #2, opened 2026-04-22
Use obsidian CLI instead of MCP


R2-t:
Hey! First and forth most, cool project.

I had investigated some obsidian MCPs before and none of them work as good as the obsidian CLI. You can provide a obsidian-cli skill such as [kepano/obsidian-cli](https://github.com/kepano/obsidian-skills/blob/main/skills/obsidian-cli/SKILL.md) and be good to go.

There's no need to overcomplicated, in my opinion.

eugeniughelbur:
Good call. Just added Obsidian CLI as the preferred method in 746a97e — it falls back to MCP, then filesystem, so nothing breaks for existing users. Thanks for the suggestion.

eugeniughelbur:
Update — turns out the 'obsidian-cli' in the Kepano repo is another Claude skill, not a terminal binary (my mistake for not checking before committing). Reverted in 48f9cfe since there's no actual CLI to detect. Happy to reopen if you're referring to a different tool I'm missing.

francislee-cn:
Hey — quick clarification: Obsidian now ships an official first-party CLI built into the desktop app (since v1.12, released April 2026). No third-party tool, MCP server, or Claude skill needed — it's bundled with Obsidian itself.

Official docs: https://obsidian.md/help/cli

eugeniughelbur:
Thanks for flagging this, and you're right, I was wrong about the CLI not existing. Just read through the v1.12 docs and yeah, `obsidian daily`, `obsidian search`, the TUI, all of it. Genuinely solid.

One thing worth drawing out though. The CLI and this skill aren't really doing the same job.

The CLI is a way to talk to Obsidian from the terminal. It runs commands you give it. Fast, well built, no AI involved.

This skill is more of an AI layer that sits on top of the vault. Most of what it does needs a language model to actually function. Things like reading a whole conversation and figuring out which parts are a person, which are a project, which are a decision, then writing linked notes across the vault automatically. Or kicking off Perplexity and Grok research runs and saving the synthesis. Or scanning the vault for contradictions between notes you wrote months apart. Those aren't commands you can call, they're judgment calls.

Where the two do overlap, like daily notes or vault search, the smart move is for the skill to call the official CLI under the hood instead of editing files directly. That's a real upgrade. Adding it to the roadmap, appreciate the nudge.

francislee-cn:
Thanks for the thoughtful reply, and great to see this making it onto the roadmap.

[...]

PR #21, opened 2026-05-20
Add per-day operation logs and supporting scripts

MPZ-00:
## What this PR does

Replaces the monolithic `log.md` with a per-day split structure under `Logs/YYYY-MM-DD.md`.
Updates `/obsidian-init` to bootstrap the new structure on fresh vaults. Ships two standalone
scripts: `migrate_log.py` (idempotent one-shot migration for existing vaults) and
`vault_stats.py` (rewrites the stats marker block in `index.md`).

Extracted from #19 — logs and supporting scripts only.

## Type of change

- [x] 📚 Documentation update

## Related issues

Related to #19 (extracts the per-day log portion)

## How I tested it

- `python scripts/migrate_log.py --vault <path> --dry-run` reports sections correctly without writing
- `python scripts/migrate_log.py --vault <path>` splits entries, skips existing files on re-run (idempotent)
- `python scripts/vault_stats.py --vault <path>` populates markers in `index.md`; re-run produces no diff
- `python scripts/vault_stats.py --vault <path> --print-only` and `--json` both work without touching files
- `/obsidian-init` step 6 now creates `Logs/` dir + pointer `log.md` + today's per-day file

## AI-first rule compliance

- [x] Per-day log files get `type: log`, `date`, `ai-first: true` frontmatter and a `## For future Claude` preamble

## Checklist

- [x] Searched existing issues and PRs before opening this one
- [x] Commit messages are descriptive
- [x] Updated `CHANGELOG.md` under "Unreleased"
- [x] Updated `SKILL.md` with per-day log structure docs


eugeniughelbur:
thanks @MPZ-00, scope is right and the script design is solid (`--dry-run`, idempotent, marker-block pattern for stats, `--print-only` and `--json` modes — all the right defaults).

two issues i'd want fixed before merge:

**1. migration regex won't match the actual log format.** the regex in `migrate_log.py` is `^##\s+(\d{4}-\d{2}-\d{2})\s*$` which expects `## YYYY-MM-DD` end-of-line. but the format the skill actually ships (per `SKILL.md`) and that real vaults use is `## [YYYY-MM-DD] action | Description` — brackets around the date, content after it. on my own vault the script would say "no `## YYYY-MM-DD` sections found, nothing to migrate." needs to be `^##\s+\[?(\d{4}-\d{2}-\d{2})\]?` or similar, and the body extraction needs to handle the trailing `action | Description` text on the same line.

**2. em-dashes in the per-day entry template.** both `PER_DAY_FRONTMATTER` and `ROOT_POINTER` use `**HH:MM** — action` with an em-dash. the repo `CLAUDE.md` is explicit: "no em-dashes in user-facing prose, use a regular dash or restructure." the template embeds an em-dash in every future log entry. needs to be ` - ` (regular dash with spaces) instead.

three more worth flagging but i'd take as follow-ups, not blockers:

- **schema mismatch in vault_stats.py.** aggregates `type: person` for People count. existing convention in real vaults uses `type: entity` for people. on my own vault stats would report 0 people. either align the canonical `references/ai-first-rules.md` to `entity` or have the script count both. worth a separate decision pass.
- **folder exclusions case-sensitive.** `EXCLUDED_FOLDERS = {"Templates", ...}` misses `templates/` lowercase, and doesn't exclude `raw/` which is the immutable source folder convention. either lowercase the comparison or expand the set.
- **other `/obsidian-*` commands not updated.** SKILL.md says "check the structure at runtime" but `/obsidian-save`, `/obsidian-log`, etc. aren't updated to actually detect old-vs-new layout. relies on the AI to read SKILL.md and make the call. fine for now but a soft contract that could drift, worth tracking.

happy to merge once the two blockers (#1 and #2) are fixed. nice work on the script structure overall.

eugeniughelbur:
merged. fast clean turnaround on both blockers:

- regex now handles `## [YYYY-MM-DD] action | description` and bare `## YYYY-MM-DD` (verified the bracket-optional + trailing-content-consumed shape against the actual log format)
- every em-dash swept from `PER_DAY_FRONTMATTER`, `ROOT_POINTER` prose, the embedded template inside the code block, and the example entry. thorough.

the three follow-ups (schema mismatch on `person` vs `entity`, lowercase folder exclusions, other `/obsidian-*` commands not yet updated to detect old-vs-new layout) — happy to track separately if you want to take them, otherwise they sit as known limitations until someone hits them.

three merged in two days. real momentum. looking forward to `/obsidian-setup` and the `/projects` + Bases discussion when you're ready.

PR #22, opened 2026-05-20
Fix vault_stats.py follow-ups and log routing in all commands

MPZ-00:
Addresses the three follow-up items flagged in the #21 review.

### `scripts/vault_stats.py`

- **People count**: now aggregates both `type: person` and `type: entity`.
  `ai-first-rules.md` uses `type: person` but real vaults often use
  `type: entity` ─ previously reported 0 people on those vaults.
- **Folder exclusions**: comparison is now case-insensitive (`part.lower()`),
  so `templates/` and `Templates/` are both excluded. Added `raw/` to the
  exclusion set (immutable source folder convention).
- **Em-dashes**: swept from `render_block` output ─ the stats block written
  to `index.md` now uses ` - ` throughout.

### Commands ─ log layout routing

Every `/obsidian-*` command that reads or appends to the operation log now
explicitly detects the vault layout instead of hardcoding `log.md`:

- **Modernized vaults** (`Logs/` exists): write `**HH:MM** - action | description`
  to `Logs/YYYY-MM-DD.md`
- **Legacy vaults**: append `## [YYYY-MM-DD] action | description` to `log.md`

Commands updated: `/obsidian-health`, `/obsidian-ingest`, `/obsidian-learn`,
`/obsidian-adr`, `/obsidian-export`, `/obsidian-reconcile`, `/obsidian-synthesize`,
`/obsidian-visualize` (append), `/obsidian-daily`, `/obsidian-synthesize`,
`/obsidian-world`, `/obsidian-learn` (read).

Also fixes two em-dashes in `commands/obsidian-init.md` and `SKILL.md` that
slipped through the per-day-logs sweep, and updates the propagation table in
`SKILL.md` to reflect the dual-path routing.

## Not in scope

- `/research` and `/x-pulse` still reference `log.md` in their command text ─
  those describe what the underlying Python scripts do and require script-level
  changes, tracked separately.
- `/obsidian-save` and `/obsidian-log` have no explicit operation log step
  (they rely on the SKILL.md propagation rule) ─ left for a follow-up
  discussion on whether explicit steps belong there.


eugeniughelbur:
merged. this is exactly the kind of follow-up PR i was hoping for. you took the entire "non-blocker, tracked as follow-up" list from the #21 review and turned it into a clean focused PR, plus the unprompted em-dash sweep in `vault_stats.py`, `obsidian-init.md` (entry template), and `SKILL.md` (format spec) - those slipped through #21's sweep and you caught them.

audit against the #21 follow-up list:

- schema mismatch (`person` vs `entity`) - fixed (counts both, with the inline comment explaining why)
- folder exclusions case-sensitivity + missing `raw/` - fixed (`part.lower()` + `raw/` added)
- 11 commands updated with the explicit `Logs/` vs `log.md` routing - fixed (soft contract from SKILL.md now hard-coded into each command's body)

the "not in scope" section is the right call - `/research`, `/x-pulse`, `/obsidian-save`, `/obsidian-log` flagged honestly as needing separate work. that kind of "here is exactly where i stopped" transparency is rare and i appreciate it.

one micro-nit, not worth changing this PR: the legacy `log.md` fallback routing instructions still embed em-dashes (e.g. `## [YYYY-MM-DD] adr | Title — decision recorded`). i think you chose backward-compat with existing legacy log.md entries on purpose - reasonable. could go em-dash-free in a future sweep when migrating users to per-day logs anyway, but leaving the legacy path alone for now keeps mixed-format files from showing up.

4 PRs from you in 2 days. looking forward to `/obsidian-setup` and `/projects` + Bases when you're ready. on `/projects` specifically there is still an open vault-structure conversation (existing `wiki/projects/` convention vs the capitalized `Projects/` your PR assumed) - happy to discuss approaches before you write code if useful, otherwise pick whichever and we'll see where it lands.

PR #20, opened 2026-05-20
Feat/sessionstart hook

MPZ-00:
## What this PR does

Adds a SessionStart hook (`hooks/load_vault_context.py`) that injects `_CLAUDE.md` into
Claude's context automatically at session start when the cwd is inside `$OBSIDIAN_VAULT_PATH`.
Eliminates the per-command re-read of `_CLAUDE.md` that burned tokens on every invocation.
Updates `scripts/setup.sh` to wire the hook alongside the existing PostCompact agent.

Extracted from #19 as requested — hook + setup wiring only.

## Type of change

- [x] ✨ New feature (non-breaking change that adds functionality)
- [x] 📚 Documentation update

## Related issues

Related to #19 (extracts the SessionStart hook portion; #19 to be closed in favour of focused PRs)

## How I tested it

- `bash scripts/setup.sh "/path/to/vault"` completes without error; `~/.claude/settings.json`
  gains a `SessionStart` entry pointing at the hook
- Hook fires on session start inside the vault: `_CLAUDE.md` content appears in context,
  no re-read needed on subsequent commands
- Hook skips silently when cwd is outside the vault path — non-vault sessions unaffected
- Path construction uses `Path(vault) / "..."` — correct separators on both Windows and POSIX

## AI-first rule compliance

N/A — this PR adds no vault-writing command.

## Checklist

- [x] Searched existing issues and PRs before opening this one
- [x] Commit messages are descriptive (focus on WHY, not just WHAT)
- [x] Updated `CHANGELOG.md` under "Unreleased"
- [x] Updated `SKILL.md` with Method 0 and the skip note for step 1
- [ ] README update — not applicable, no new user-facing command

PR #27, opened 2026-05-23
Fix NotADirectoryError in check_empty_folders for vaults with root-level files

napstor:
## Summary

`vault_health.py:check_empty_folders` crashes with `NotADirectoryError` on any vault that has files at its root — including the `_CLAUDE.md` / `index.md` / `log.md` trio that `/obsidian-init` generates itself.

## Repro

```bash
python scripts/vault_health.py --path ~/path/to/vault --json
```

Observed on macOS APFS, vault containing `_CLAUDE.md`, `index.md`, `log.md` at root (standard `/obsidian-init` output):

```
Traceback (most recent call last):
  File ".../vault_health.py", line 321, in <module>
    main()
  File ".../vault_health.py", line 312, in main
    result = run_health_check(vault)
  File ".../vault_health.py", line 243, in run_health_check
    ("Empty folders", check_empty_folders(vault)),
  File ".../vault_health.py", line 176, in check_empty_folders
    if not list(folder.iterdir()):
NotADirectoryError: [Errno 20] Not a directory: '.../_CLAUDE.md'
```

## Root cause

`vault.rglob("*/")` returns entries that aren't always directories on some filesystems — the trailing slash is not a strict directory filter. `folder.iterdir()` then raises `NotADirectoryError` on the first non-directory entry, killing the whole health check.

## Fix

Single-line defensive guard:

```python
if not folder.is_dir():
    continue
```

inserted before `folder.iterdir()`. Matches the same pattern already used in `check_orphans` (line 208 of the same file).

## Test plan

- [x] Reproduce the crash on a vault containing `_CLAUDE.md`/`index.md`/`log.md` at root → `NotADirectoryError`
- [x] Apply the patch → `vault_health.py` runs to completion, emits valid JSON
- [x] Other checks (`check_broken_links`, `check_orphans`, frontmatter checks) unaffected
- [x] No new dependencies, no behavior change for vaults that previously worked

### DIFF
--- a/scripts/vault_health.py
+++ b/scripts/vault_health.py
@@ -173,6 +173,8 @@ def check_empty_folders(vault: Path) -> list:
     for folder in vault.rglob("*/"):
         if any(p in EXCLUDE_DIRS for p in folder.parts):
             continue
+        if not folder.is_dir():
+            continue
         if not list(folder.iterdir()):
             rel = str(folder.relative_to(vault))
             issues.append({


eugeniughelbur:
merged. defensive `is_dir()` guard is exactly the right shape, matching the pattern already used in `check_orphans` at line 208. solid PR hygiene too - traceback + root cause + test plan in the description meant i didn't have to re-derive any of it.

worth noting: this is the third external bug fix landed on this repo in two weeks (@jameswolensky in the US on PR #17, @MPZ-00 in Germany across PRs #20-25, now you in Cyprus on this one). three different contributors, three different bugs, all caught in real usage. that pattern is the actual product-market-fit signal for an open source skill, stronger than stars or forks.

thanks for adding to it.

PR #95, opened 2026-07-07
Fix heal_links.py import crash and resolve Title-case wikilinks (slug matcher + batch mode)

liauw-media:
## What this fixes

`scripts/heal_links.py` currently crashes on import, and once that is fixed it still cannot resolve the most common broken-link shape in a real vault.

### 1. Import crash
It imports `check_broken_links` from `vault_health`, but that function was removed during the resolver refactors (the module now exposes `check_wanted_notes`). Every run fails immediately:

```
ImportError: cannot import name 'check_broken_links' from 'vault_health'
```

Its message regex also still expects the old `Broken link [[...]]` wording, while `check_wanted_notes` emits `[[...]] - wanted by <rel>`.

### 2. Title-case links were never resolved
Even with the import fixed, the matcher could not repoint the dominant real case: a note links a sibling by its human Title (`[[Host iptables rules ...]]`) while the file on disk is kebab-cased (`host-iptables-rules-....md`). The old matcher compared a space-title against a hyphen-stem, so those links never matched and stayed "wanted" forever.

## The fix

- Rewire to `check_wanted_notes` and update the message regex, so the healer's score is exactly the health check's `wanted_note` count again.
- Add a **slug matcher**: lowercase both sides and collapse every run of non-alphanumeric characters to a single space, so `Title Case`, `kebab-case`, punctuation, and stray slashes like `/24` all compare equal. Matching is deterministic (exact slug); the existing `difflib` pass is kept only as a last resort. Ambiguous slugs (2+ candidate notes) are left for triage, never auto-picked.
- Add a **`--batch` mode**: the original loop reloads the whole vault and re-runs the wanted-note scan after every single fix (O(n^2) file reads), impractical for thousands of links. `--batch` collects all unambiguous fixes in one pass, applies them grouped per file, and recounts once.
- **Preserve readability**: rewrite `[[Title]]` to `[[kebab-basename|Title]]` (display alias) instead of `[[kebab-basename]]`, so healed links render identically.

## Testing

- `uv run pytest -q` -> **29 passed**.
- End-to-end on a real 201-note vault:

```
python scripts/heal_links.py --path <vault> --dry-run    # 1305 safe, 0 ambiguous
python scripts/heal_links.py --path <vault> --batch
  wanted links before:     1962
  safe auto-fixes applied:  1715 (across 155 files)
  wanted links after:        249   # remainder are genuine stubs / non-existent targets
```

As a side effect `vault_health` orphans dropped 42 -> 7 (healed links give previously-unreferenced notes an incoming link).

## Compatibility

Backward compatible: `--dry-run` and `--apply --max N` behave as before; `--batch` is additive.

## Note

I scoped this PR to the code change. Happy to add a `CHANGELOG.md` entry under `Unreleased > Fixed` if you'd like one in the same PR - just say the word.

### DIFF
--- a/scripts/heal_links.py
+++ b/scripts/heal_links.py
@@ -1,40 +1,53 @@
 #!/usr/bin/env python3
 """
-heal_links.py - a closed loop that heals broken wikilinks in an Obsidian vault.
+heal_links.py - heals broken/"wanted" wikilinks in an Obsidian vault.
 
-The loop (this is the whole idea):
-    start once, then ON ITS OWN it repeats:
-        find a broken link
-        decide the fix  (a plain name match - no AI, no API key, no cost)
-        write the fix into the note
-        run the broken-link check again  ->  a countable score
-        repeat until nothing safe is left, the score stops dropping, or the budget runs out
+A wiki-style vault often links a note by its human Title ([[Host iptables rules ...]])
+while the file on disk is kebab-cased (host-iptables-rules-....md). Those links do not
+resolve, so vault_health counts them as "wanted notes". This script repoints each such
+link to the real note when exactly one note is an unambiguous match, preserving the
+original text as a display alias: [[kebab-basename|Host iptables rules ...]].
 
-This file IS the loop. vault_health's check is the score. You press start one time.
-It is "closed" on purpose: bounded, safe matches only, a hard recount every pass.
+Matching is deterministic and safe:
+  1. exact stem/alias match,
+  2. slug match - lowercase both sides and collapse every run of non-alphanumeric
+     characters to a single space, so "Title Case", "kebab-case", punctuation and
+     stray slashes all compare equal (this is what resolves the Title<->kebab case),
+  3. a last-resort fuzzy match (difflib) only when exactly one close name exists.
+Ambiguous links (two or more matches) and links with no match are counted and left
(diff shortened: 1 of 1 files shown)


eugeniughelbur:
Merged, thank you. The import fix and the deterministic slug matcher are exactly right, and confirmed `heal_links.py` was crashing on import before this.

I added one safety follow-up on top (c9874c1) before landing it. Testing `--dry-run` against a real vault, the difflib fuzzy fallback was auto-classifying single close-matches as `easy_fix`, and at cutoff 0.84 that produced confident-but-wrong repoints: `[[Google Ads]]` -> `Google Docs`, and three distinct `[[ADR-2026-05-*]]` records all collapsing onto `ADR-2026-05-11`. `--batch` would have applied those unattended.

Change: fuzzy hits now route to `ask_claude` (triage) and never auto-apply, so only exact and slug matches are healed automatically. I also made `slugify` fold diacritics (NFKD + ASCII), so accent-only cases like `Ivan` -> `Iván` stay certain slug matches instead of falling to the guesser. On the test vault that took auto-fixes from 49 (40 fuzzy, several wrong) down to 7 unique links, all verified correct; the fuzzy cases now surface under "left for AI triage". 29 tests still pass.

PR #97, opened 2026-07-09
feat: add Tavily search source to free research aggregator

manisrinivasan2k1:
## Summary

Adds `TavilySource` as an optional, additive search provider in the free-source research aggregator. When `TAVILY_API_KEY` is set, Tavily participates alongside DuckDuckGo and other free sources in the parallel `ThreadPoolExecutor` aggregation, improving result quality and reliability.

DuckDuckGo remains the zero-key baseline -- Tavily is purely additive.

## Changes

- **New file: `scripts/research/lib/sources/tavily.py`** -- `TavilySource` class implementing the `SourceClient` protocol (`name='tavily'`, `search(query, n) -> list[Result]`), wrapping `tavily-python`'s `TavilyClient.search()`.
- **`scripts/research/lib/sources/__init__.py`** -- Export `TavilySource`.
- **`scripts/research/research.py`** -- In `_free_sources()`, conditionally append `TavilySource()` when `TAVILY_API_KEY` env var is set.
- **`tests/test_research_sources.py`** -- Added `test_tavily_source_search` with mocked SDK calls.
- **`pyproject.toml`** -- Added `tavily-python>=0.3.0` dependency.
- **`.gitignore`** -- Added `.pytest_cache/`.

## Dependency changes

- Added `tavily-python>=0.3.0` to `pyproject.toml`

## Environment variable changes

- `TAVILY_API_KEY` (optional) -- when set, enables Tavily in the free-source aggregator

## Notes for reviewers

- No existing behavior changes -- DuckDuckGo and all other sources are untouched
- All 30 tests pass

### Automated Review

- Passed after 2 attempt(s)
- Final review: The implementation is correct and addresses all 4 issues from the first review. TavilySource is added as an optional, additive source in the free aggregator — DuckDuckGo is preserved. The `=0.3.0` file is absent from the git tree. `scripts/research/lib/sources/__init__.py` is empty (blob hash e69de29b). `pyproject.toml` has the tightened `tavily-python>=0.5.0` bound. The test file adds both an import smoke test (TavilySource in the names set, count bumped to 11) and a proper mock-based functional test. uv.lock is updated. Two minor points: TAVILY_API_KEY is not documented in .env.example, and an unrelated .pytest_cache/ line was added to .gitignore. Neither blocks approval.


eugeniughelbur:
Thanks for this - the implementation is clean and well-tested, and the lazy import + key-gated activation are exactly right.

One change before I can merge: please make `tavily-python` an optional dependency rather than a hard one. The free-source aggregator's core promise is that it runs with zero keys and zero paid dependencies, so I do not want to pull `tavily-python` (and its transitive deps) into every install when only key-holders use it. Your lazy imports already support this cleanly.

Concretely: move it out of the main `dependencies` list into an optional extra, e.g.

    [project.optional-dependencies]
    tavily = ["tavily-python>=0.5.0"]

and keep the runtime `from .lib.sources.tavily import TavilySource` guarded so a missing package degrades gracefully (skip Tavily with a note) instead of raising ImportError when the key is set but the extra is not installed.

The unrelated `.pytest_cache/` gitignore line is fine to leave in. Ping me when updated and I will re-run and merge.

eugeniughelbur (changes requested):
thanks for this - the shape is right (env-gated, additive, DuckDuckGo stays the zero-key baseline), and since /research-deep reuses `_free_sources` your one edit point covers both commands.

three things before it can land:

1. rebase on main. this predates a 24-PR sprint that rewrote `tests/test_research_sources.py`, `pyproject.toml` and `uv.lock`, so it's conflicting now (the suite went 30 -> 129 tests in the meantime)
2. drop the `tavily-python` dependency. every other source hits its API with plain `requests`, and tavily search is a single POST to `https://api.tavily.com/search`. keeps the dep tree lean for everyone who never sets the key
3. one doc line. free mode is documented as key-less in `commands/research.md` and the README, so an optional keyed booster needs a sentence there, plus `TAVILY_API_KEY` in the recognized-env-vars list in `scripts/research/lib/source_config.py`'s docstring

happy to help with the rebase if you get stuck.

eugeniughelbur:
The feature is now on main via #145 (which auto-closed this) - with you credited as co-author on the merge commit. Since this branch had been quiet since the review two weeks back and had drifted behind main, I reimplemented it there rather than leave it hanging.

What changed from this PR, per the review notes: plain `requests` against the REST API instead of the `tavily-python` SDK (dependency list unchanged), Bearer-header auth so the key never travels in the request body, the shared session + disk cache every other source uses, and an explicit key-less guarantee in the docs (`TAVILY_API_KEY` optional, the free pool never requires it).

Thanks for proposing this and for the original implementation, @manisrinivasan2k1 - the wiring shape (gate on the env var inside `_free_sources`) is yours.

PR #125, opened 2026-07-11
Fix /research and /research-deep ignoring PERPLEXITY_API_KEY from .env

### LINKED ISSUE #124: /research and /research-deep silently ignore PERPLEXITY_API_KEY from .env (env checked before load_dotenv runs)
## Summary

`/research` and `/research-deep` always fall back to **free-sources mode** even when a valid `PERPLEXITY_API_KEY` is configured the documented way (in `~/.config/obsidian-second-brain/.env`). The fallback is silent, so paid-mode users get the free pipeline without any indication that their key was ignored.

## Root cause

In `scripts/research/research.py` and `scripts/research/research_deep.py`, `main()` decides free vs paid by reading the process environment directly:

```python
use_free = force_free or not os.environ.get("PERPLEXITY_API_KEY", "").strip()
```

At this point nothing has loaded the `.env` file yet — `load_dotenv(ENV_PATH)` only runs when `scripts/research/lib/config.py` is first imported, and both scripts import `lib.*` modules lazily *inside* `run_free()` / `run_paid()`, i.e. only **after** the decision has been made. Since the documented setup puts keys only in the `.env` file (not the shell environment), the check always sees an empty environment and picks free mode.

`podcast_extract.py` is not affected: it imports `from .lib import grok, podcast, vault` at module top, which pulls in `config.py` (and thus `load_dotenv`) before anything reads the environment.

## Reproduction

1. Configure `PERPLEXITY_API_KEY` in `~/.config/obsidian-second-brain/.env` (per install.sh / README), and make sure it is **not** exported in your shell.
2. Run `uv run -m scripts.research.research "any topic"`.
3. Observe the free-sources JSON pipeline runs instead of the Perplexity paid flow.

## Fix

Load the shared `.env` before the decision. Note that importing `lib.config` for this would be wrong: `config.py` calls `get_required("OBSIDIAN_VAULT_PATH")` at import time, which would break zero-config free mode (which `lib/source_config.py` explicitly protects: *"they must be importable with no vault configured at all"*). Importing `lib.source_config` instead loads the same `.env` in a guarded, never-fails block:

```diff
     if not topic:
         print("Usage: /research <topic> [--free] [--academic]", file=sys.stderr)
         return 2

+    # Load the shared ~/.config/obsidian-second-brain/.env before deciding
+    # free vs paid. source_config does this without requiring vault config
+    # (unlike lib.config), so zero-config free mode keeps working.
+    from .lib import source_config  # noqa: F401
+
     use_free = force_free or not os.environ.get("PERPLEXITY_API_KEY", "").strip()
     if use_free:
         return run_free(topic, academic)
     return run_paid(topic)
```

Same change in `research_deep.py`'s `main()`.

Verified locally: with the key only in `.env`, paid mode is now selected; with a fake `$HOME` (no `.env`, no vault config), free mode still works and nothing raises. Full test suite passes (119 passed).

MichaelHabermas:
Fixes #124

### DIFF
--- a/scripts/research/research.py
+++ b/scripts/research/research.py
@@ -190,6 +190,11 @@ def main(argv: list[str]) -> int:
         print("Usage: /research <topic> [--free] [--academic]", file=sys.stderr)
         return 2
 
+    # Load the shared ~/.config/obsidian-second-brain/.env before deciding
+    # free vs paid. source_config does this without requiring vault config
+    # (unlike lib.config), so zero-config free mode keeps working.
+    from .lib import source_config  # noqa: F401
+
     use_free = force_free or not os.environ.get("PERPLEXITY_API_KEY", "").strip()
     if use_free:
         return run_free(topic, academic)
--- a/scripts/research/research_deep.py
+++ b/scripts/research/research_deep.py
@@ -357,6 +357,11 @@ def main(argv: list[str]) -> int:
         print("Usage: /research-deep <topic> [--free] [--academic]", file=sys.stderr)
         return 2
 
+    # Load the shared ~/.config/obsidian-second-brain/.env before deciding
+    # free vs paid. source_config does this without requiring vault config
+    # (unlike lib.config), so zero-config free mode keeps working.
+    from .lib import source_config  # noqa: F401
+
     use_free = force_free or not os.environ.get("PERPLEXITY_API_KEY", "").strip()
     if use_free:
         return run_free_deep(topic, academic)


eugeniughelbur:
merged, thanks! the root-cause writeup was spot on - lazy lib imports meant the free/paid decision ran before anything loaded the .env. made it easy to verify. added a smoke test on main (5647f40) so it stays fixed.

appreciate it!

Issue #79, opened 2026-06-26
Roadmap: native Hermes curator mode (beyond the MCP connector)


eugeniughelbur:
Splitting this out of #60 so the connector work and the deeper "Hermes runs the vault" work are tracked separately, per @bkmaibach's suggestion.

## Where we are

- **Phase 0 - run on Hermes (done).** The skill is model-agnostic; the OpenCode / Codex / Gemini builds run on Hermes via OpenRouter or local Ollama. Docs shipped.
- **Phase 1 - MCP connector (done, #77).** A stdio MCP server exposes the vault as 6 tools (`obsidian_search`, `obsidian_read_note`, `obsidian_save_note`, `obsidian_capture`, `obsidian_list_skills`, `obsidian_get_skill`). This is the bounded "Hermes uses the vault as a knowledge layer" doorway. It is deliberately a connector, not the end-state.

## Phase 2 - native Hermes curator mode (this issue)

Hermes natively running obsidian-second-brain as its knowledge curator, using its own native primitives instead of routing everything through one MCP process:

- [x] **Hermes skill adapter** - a new adapter target (`adapters/hermes/`) that emits the command playbooks as native Hermes skills, the same way the Codex adapter now emits native Codex Agent Skills (#78). Progressive disclosure, in-session execution.
- [x] **Richer MCP tools** - safe updates to *existing* notes (not just create), frontmatter / link validation, backlink queries, and a `vault_health` tool. The current set is read + append-only by design; curator mode needs guarded mutation.
- [x] **Hermes cron** replacing the Claude Code scheduled agents (nightly synthesis, reconcile, health) with Hermes-native cron jobs.
- [x] **Hermes lifecycle hook / background review** replacing the PostCompact hook, so vault maintenance fires on Hermes's own session lifecycle.

## Open design question

(Carried from #60.) The richer mutation tools are the riskiest piece - they touch existing notes. Likely they stay as *skills the agent executes* (multi-step, reviewable) with the MCP layer providing only the bounded primitives, rather than a single "rewrite note" function call. Input welcome.

Co-design / test seats open. /cc @bkmaibach @savez

eugeniughelbur:
[...]

PR #135, opened 2026-07-16
Fix Hermes install story: blueprints never arm on install (#134)

### LINKED ISSUE #134: Hermes dist: INSTALL.md/HOOKS.md arming story doesn't match Hermes - blueprints never arm on install, and the hook config/headless entrypoint are wrong
## What's wrong

Verified against the official Hermes docs (hermes-agent.nousresearch.com) and a live Hermes CLI while installing the scheduled agents for real. Four claims in the Hermes build are incorrect:

1. **`dist/hermes/INSTALL.md` points at a directory Hermes never reads.** It says to `cp -R dist/hermes/optional-skills/. ~/.hermes/optional-skills/`. There is no `~/.hermes/optional-skills/`; installed skills live in `~/.hermes/skills/` ("the primary directory and source of truth"), discovered as `<category>/<name>/SKILL.md`.

2. **"a Hermes blueprint arms as soon as its skill is loaded" (`HOOKS.md`) is the opposite of Hermes's contract.** The Automation Blueprints docs state: "Blueprints never schedule anything silently - you always confirm before the job is created." `hermes skills install` registers a blueprint as a *suggested* cron job (accepted via `/suggestions`); a manually copied skill registers nothing. Following our docs leaves the user with zero scheduled jobs and no error - reproduced on a live install: skills detected by `hermes skills list`, `/blueprint obsidian-nightly` finds nothing, nothing armed. The working path is `hermes cron create "<schedule>" "<prompt>" --skill <name> --workdir <vault>` (verified: creates and runs the jobs).

3. **Hooks are declared in `~/.hermes/config.yaml`, not `cli-config.yaml`.** `hooks/hermes-hooks.cli-config.example.yaml` and both docs reference a config file that doesn't exist.

4. **The headless entrypoint `hermes run --quiet` doesn't exist, and the hook passes the prompt on stdin.** The documented one-shot mode is `hermes -z PROMPT` (or `hermes chat -q QUERY -Q`), which takes the prompt as an *argument*. `hooks/obsidian-hermes-session-end.sh` defaults to `hermes run --quiet` and pipes the prompt via stdin, so the `on_session_end` consolidation silently does nothing even when opted in. This also resolves the "one unverified seam" flagged in HOOKS.md (#79, Phase 2).

## Where it lives

All of this is generated by `adapters/hermes/adapter.sh` (INSTALL.md + HOOKS.md heredocs and the blueprint SKILL.md wrapper text, which repeats the "install to arm" claim), plus the two shipped hook files. The blueprint frontmatter itself (`metadata.hermes.blueprint` with `schedule`/`deliver`/`prompt`/`no_agent`) matches the documented schema exactly - only the install/arming story around it is wrong.

## Fix

PR incoming: correct the arming instructions (cp into `~/.hermes/skills/`, arm via `hermes cron create` or `/suggestions` after a registry install), retarget the hook config to `~/.hermes/config.yaml`, switch the hook default to `hermes -z` with argument-passed prompt, and update the smoke test.

Litash:
Closes #134.

## What this fixes

The Hermes build's scheduled-agent install instructions described a mechanism Hermes doesn't have. Following them left a user with detected skills but **zero armed schedules and no error** - reproduced on a live Hermes install. Verified against the official docs (hermes-agent.nousresearch.com: Automation Blueprints Catalog, Creating Skills, Cron, Hooks, CLI reference) and a live Hermes CLI where the corrected `hermes cron create` flow was exercised end-to-end (all four jobs created and listed).

Four corrections, all in the adapter-emitted docs and the two shipped hook files:

1. **Install destination.** `~/.hermes/optional-skills/` (never read by Hermes) → `~/.hermes/skills/` (documented "primary directory and source of truth"; the nested `<category>/<name>/SKILL.md` layout is discovered fine).
2. **Arming.** "a Hermes blueprint arms as soon as its skill is loaded" via `hermes skills install <name>` → explicit arming, matching Hermes's contract ("Blueprints never schedule anything silently"): `hermes cron create "<schedule>" "<prompt>" --skill <name> --workdir <vault>`, or accepting the suggested job from `/suggestions` after a registry install. INSTALL.md, HOOKS.md, and the blueprint SKILL.md wrapper text all updated; the blueprint frontmatter itself (`metadata.hermes.blueprint`) already matched the documented schema and is unchanged.
3. **Hook config.** `cli-config.yaml` doesn't exist; shell hooks are declared under `hooks:` in `~/.hermes/config.yaml`. Template renamed to `hooks/hermes-hooks.config.example.yaml`.
4. **Headless entrypoint.** `hermes run --quiet` doesn't exist and the prompt was piped on stdin, so the opt-in `on_session_end` consolidation silently did nothing. Default is now `hermes -z` (documented one-shot mode) with the prompt passed as the final argument - closing the "one unverified seam" HOOKS.md flagged from #79 Phase 2.

## Testing

- `bash scripts/build.sh --platform hermes` - clean; grep confirms no stale claims (`cli-config`, `arms as soon`, `hermes run --quiet`, `~/.hermes/optional-skills`) survive in `dist/hermes/`.
- `uv run pytest -q` - 164 passed. The hermes smoke test now also asserts the emitted docs teach `hermes cron create` and never mention the phantom directory or auto-arming.
- `bash -n` on the modified hook script.
- The corrected user flow (cp into `~/.hermes/skills/`, `hermes cron create` per agent, `hermes cron list`) was run for real on a live Hermes; all four jobs armed and scheduled.

Co-authored with Claude.


eugeniughelbur:
Merged into main via 039f8a3 (resolved a CHANGELOG conflict against the recent bg-agent work, kept both entries). Your commit is preserved in history. Excellent fix - verified-against-live-Hermes across all four bugs. Thank you @Litash.

PR #140, opened 2026-07-18
feat(scripts): vault_health.py — add user-level .vault-config.json excludes

Rongles-World:
## Motivation

The current `vault_health.py` has hardcoded `EXCLUDE_DIRS` (9 items) that cannot be customized per-vault. Academic / research vaults typically have directories that should be excluded from health scans (e.g., distillation pool atomic cards, backup snapshots, academic transcription directories). Adding these to the hardcoded list makes the script harder to maintain; not adding them produces ~10,000+ false-positive noise issues.

This PR adds a user-level config file (`.vault-config.json`) that lets users extend the exclude list without modifying the script.

## Changes

- Add `load_vault_config(vault)` function — reads `<vault>/.vault-config.json`
- Add `_USER_EXCLUDES` module-level dict — populated per scan
- Add `_should_skip_path()` and `_should_skip_path_file_index()` helpers
- Wire up in `run_health_check()` — load config before scanning
- Replace 3 inline checks in `index_vault_files`, `load_vault`, `check_empty_folders`

## Schema

```json
{
  "exclude-dirs":  ["_distillation-pool", "_candidates", ...],
  "exclude-paths": ["Archive/Backup", ...]
}
```

- `exclude-dirs`: list of dir names (matched as path components)
- `exclude-paths`: list of path prefixes (matched as vault-relative POSIX paths)
- Both are optional and **additive** — hardcoded `EXCLUDE_DIRS` always applies

## Backward Compatibility

- Hardcoded `EXCLUDE_DIRS` unchanged
- Missing `.vault-config.json` → silently skip (no error)
- Malformed JSON → silently skip (no error)
- Empty config → behaves exactly as before
- All 172 existing tests still pass

## Validation

Tested on a real-world 12,310-note research vault:
- Before whitelist: 15,045 issues
- After whitelist (with `.vault-config.json`): 2,321 issues
- **Reduction: 84.6%** (12,724 false-positive noise eliminated)
- Scan time: 5+ minutes → 19 seconds (93.7% faster)

## Zero New Dependencies

Pure stdlib (uses existing `json` + `pathlib` imports). No `pyproject.toml` changes needed.

## New Tests

Added `tests/test_user_excludes.py` (4 tests):
- `test_user_exclude_dirs_suppress_orphans`
- `test_user_exclude_paths_suppress_frontmatter_gap`
- `test_missing_config_is_silently_ignored`
- `test_malformed_config_is_silently_ignored`

### DIFF
--- a/scripts/vault_health.py
+++ b/scripts/vault_health.py
@@ -40,6 +40,58 @@ EXCLUDE_DIRS = {
 }
 FILE_INDEX_EXCLUDE_DIRS = EXCLUDE_DIRS - {"Templates"}
 EXCLUDE_ROOT_FILES = {"AGENTS.md", "INSTALL.md"}
+
+
+# User-configurable excludes loaded from `<vault>/.vault-config.json`.
+# Schema:
+#   {
+#     "exclude-dirs":  ["_蒸馏池_书稿", "_candidates", ...],   # merged with EXCLUDE_DIRS
+#     "exclude-paths": ["Archive_归档/04_备份", ...]            # path-prefix exclusions
+#   }
+# `exclude-dirs` adds dir names to the skip check; `exclude-paths` matches any
+# vault-relative path that starts with one of the prefixes (POSIX separator).
+# Both lists are optional and additive - the hardcoded EXCLUDE_DIRS always applies.
+_USER_EXCLUDES: dict = {"dirs": set(), "paths": []}
+
+
+def load_vault_config(vault: Path) -> dict:
+    cfg_path = vault / ".vault-config.json"
+    out = {"dirs": set(), "paths": []}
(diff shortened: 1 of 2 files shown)


eugeniughelbur:
Thanks for this @Rongles-World, the per-vault `.vault-config.json` idea is a real gap on large vaults. I saw you closed it without a note, so I've reimplemented the feature and opened #142 with credit to you.

Two small changes from your version: the excludes are threaded through as an optional object rather than a per-run module global, and user `exclude-paths` are kept out of the link-resolution index so a live link into an excluded path still resolves. Same config schema and behavior otherwise, plus a test that the hardcoded excludes can never be weakened by a config.

If you'd meant to rework it yourself, happy to defer, just say so on #142.

Issue #139, opened 2026-07-18
Antigravity does not discover the gemini-cli build; proposal: unified agent-skills adapter installed via skills.sh


Litash:
## Problem

Google has replaced Gemini CLI with **Antigravity**. Installing the current `dist/gemini-cli/` build into a workspace gives Antigravity users nothing: Antigravity still parses `GEMINI.md` / `AGENTS.md` as passive context files, but its native skill discovery only scans `<workspace>/.agents/skills/<name>/SKILL.md` (plus its global skill directories). The gemini-cli build ships a `GEMINI.md` routing table and command bodies under `.gemini/commands/`, so nothing lands where Antigravity looks and no skills are detected.

Google's own migration doc confirms this directly: workspace skills must be relocated from `.gemini/skills/` to `.agents/skills/` "for the Antigravity agent to recognize them as active slash commands."

## Verified platform facts (checked against official docs, 2026-07-17)

- **Antigravity**: workspace skills at `.agents/skills/<folder>/SKILL.md`; frontmatter `description` required, `name` optional (defaults to folder name); progressive disclosure (name + description always visible, body lazy-loaded); the CLI also surfaces each workspace skill as a slash command (`/skills`). Global rules live at `~/.gemini/GEMINI.md`.
- **Codex CLI**: native Agent Skills at the same `.agents/skills/` path (what `adapters/codex-cli` already emits).
- **OpenCode**: also reads `.agents/skills/` (project) and `~/.agents/skills/` (global); validates frontmatter strictly - only `name`, `description`, `license`, `compatibility`, `metadata` (string map) are accepted, and `name` must match the directory name.
- **skills.sh** (`npx skills`, vercel-labs/skills): GitHub-as-registry installer for Agent Skills covering 70+ agents including codex, opencode, and antigravity. A project-scope install writes one shared `.agents/skills/` tree that all three harnesses read.

## Suggested solution: one unified `agent-skills` adapter instead of a fourth bespoke one

The obvious fix is a new `adapters/antigravity/` adapter, but that adds a fourth near-duplicate build while Codex, OpenCode, and Antigravity have all converged on the same open Agent Skills standard and the same `.agents/skills/` path. The better move seems to be:

1. **Add `adapters/agent-skills/adapter.sh`** emitting one spec-compliant tree at `dist/agent-skills/`:
[...]

PR #141, opened 2026-07-18
Add agent-skills adapter: unified .agents/skills/ build for Antigravity/Codex/OpenCode (#139)

eugeniughelbur:
Closes the discovery gap reported in #139: Google Antigravity only detects skills under `<workspace>/.agents/skills/<name>/SKILL.md`, so a user who installed the `gemini-cli` build got nothing (its `GEMINI.md` is read only as passive context). Codex, OpenCode, and Antigravity have all converged on that same open Agent Skills path, so this adds **one** adapter that emits a single spec-compliant tree all three read, rather than a fourth near-duplicate.

Direction and prototype proposed by @Litash in #139. This lands the **additive core** so it isn't blocked; see "Deliberately deferred" below.

## What this adds

- **`adapters/agent-skills/adapter.sh`** builds `dist/agent-skills/skills/<name>/SKILL.md` (43 command skills; `/obsidian-calendar` stays Claude-only) plus a shared **`obsidian-core`** skill carrying `references/`, `scripts/`, and `pyproject.toml`. The command skills' `uv run --directory SKILL_ROOT ...` invocations are rewritten to resolve against `.agents/skills/obsidian-core`.
- **Spec-minimal frontmatter** (`name`, `description`, string-map `metadata.category`) so OpenCode's strict validator accepts it and each `name` matches its directory. No root `SKILL.md` (it would shadow the nested skills in skills.sh discovery).
- **Self-sufficient skills**: a vault-root resolution preamble (`$OBSIDIAN_VAULT_PATH` else the working directory) and the embedded AI-first write spec, so no session-start hook is required and the spec survives a partial install.
- **`trigger-mode: proactive`** (new optional source key) encodes the selection policy into each description, the only signal these harnesses use for implicit selection: capture-type commands (save, capture, log, person, decide, daily, task) read "use proactively", everything else "use only when explicitly asked".
- **`INSTALL.md`** (skills.sh path + manual `cp -R` fallback + per-harness notes, including Antigravity's headless `--add-dir` requirement) and **`global-rule-snippet.md`** for the optional always-on vault-routing rule.
- Docs: `build.sh --help`, README install section + counts, SKILL.md, CHANGELOG.

## Deliberately deferred (not in this PR)

- **No deprecations.** `codex-cli`, `gemini-cli`, and `opencode` builds are unchanged. Retiring the two overlapping adapters is reversible risk I'd rather not bundle with a new-build change, and I want the unified build proven in a release first. Clean follow-up PR.
- **Docs framing:** I described Antigravity as *differing* from Gemini CLI discovery rather than "replacing" it, since Gemini CLI still exists.

## Verification

- `bash scripts/build.sh --platform agent-skills` -> 44 skills (43 + obsidian-core), calendar excluded, no root SKILL.md.
- All 44 SKILL.md pass OpenCode-strict frontmatter checks (name matches dir, spec-minimal keys, string-map metadata).
- New `tests/test_smoke.py::test_agent_skills_build_generates_spec_compliant_tree`; full suite **169 passed**.

**What I could not test here** (would love @Litash to confirm on this branch, since their prototype already did): live skill discovery on **OpenCode** and the **Antigravity CLI** (`--add-dir` headless + interactive `/skills`). Their branch verified these on OpenCode 1.17.15 and Antigravity headless; this PR's checks are mechanical (build shape + strict-frontmatter + smoke test).


eugeniughelbur:
Post-merge live verification (the gap called out in the PR description) - done on a clean environment:

- **skills.sh discovery**: `npx skills add ./dist/agent-skills --list` finds exactly 44 skills (43 commands + obsidian-core); descriptions incl. the trigger policy come through intact.
- **Real install**: `npx skills add ... -a codex -a opencode -a antigravity` from a fresh workspace writes one shared `.agents/skills/` tree serving all three.
- **Engine works from the installed tree**: `uv run --directory ".agents/skills/obsidian-core" scripts/vault_health.py --path <vault> --json` returns clean JSON, and module-style `-m scripts.research.research` resolves - dependencies come from the shipped pyproject, no extra install step.
- **Live OpenCode (strict validator)**: `opencode debug skill` in an isolated HOME loads **all 44 skills** from the workspace `.agents/skills/` with **zero validation warnings**; proactive/explicit trigger policies and the embedded AI-first spec survive parsing.

Antigravity reads the identical path/spec OpenCode just accepted and was live-tested on @Litash's fork of this design, so calling it verified.

One dev-machine-only caveat worth noting for contributors: if the repo is symlinked into `~/.claude/skills/`, OpenCode also discovers the gitignored `dist/codex-cli/.agents/skills/` tree inside it and skill names shadow each other across locations. Normal installs are unaffected (dist/ never ships).

Litash:
## Verification: confirmed live on OpenCode + Antigravity (this PR's merged `04d540e` build)

@eugeniughelbur - ran the checks you flagged, against this PR's actual built tree, not the #139 prototype. Rebuilt `04d540e` in a clean worktree (`bash scripts/build.sh --platform agent-skills` -> 44 skills, no root SKILL.md), installed the tree into a throwaway workspace's `.agents/skills/`, and drove both harnesses. Tools: `agy` 1.1.4, `opencode` 1.17.15.

| Check | Result |
|---|---|
| **OpenCode** discovery + strict frontmatter validation | all 44 skills loaded, **0** warnings/errors |
| **Antigravity CLI** headless (`agy --add-dir <ws> --print`) | all 44 discovered |
| **Antigravity** implicit selection (proactive) | "we just decided to go with PostgreSQL over SQLite for the side project, keep this for later" -> selects `obsidian-decide` and says it would fire proactively, citing the proactive wording in the description |
| **Antigravity** explicit-only refusal | idle chat about vector databases -> declines to auto-run `research`, citing its "only when explicitly asked" wording |
| **Antigravity** interactive `/skills` palette | all 44 render as browsable slash commands with descriptions (the two above the fold, `create-command` and `idea-discovery`, scroll into view; they sort before `notebooklm`) |

So the `trigger-mode` policy isn't just discovered - it actually steers selection in both directions - and the interactive palette surface renders the full set. The discovery gap from #139 is closed end-to-end on both harnesses.

One note on scope: I did not separately drive **Codex CLI**, but it reads the identical `.agents/skills/` path with the identical spec-minimal frontmatter that OpenCode's strict validator already accepted here, so the risk there is low.

Two cosmetic nits spotted during verification (non-blocking, candidates for the follow-up you mentioned):

- **Run-on descriptions** where the source has no trailing period - e.g. `obsidian-save` renders "...worth keeping from this conversation to the vault Triggers: save this...". Inserting a period before the folded-in `Triggers:` / `Use ...` sentence reads cleaner.
- **`obsidian-core` appears in the `/skills` palette** as a selectable command even though its own description says "support files only ... do not invoke directly." Harmless, just a slightly odd entry when browsing.

Happy to open a small follow-up PR for the deprecation banners (codex-cli / opencode) plus the contributor-doc updates - `CLAUDE.md` and `architecture.md` still describe the pre-agent-skills adapter set ("the other five adapters ... routing table") - if that's useful. I have those staged from the prototype branch.

Litash:
Opened the follow-up I mentioned: #146. Scoped to exactly what was deferred here, no overlap with this PR:

- deprecation banners on the `codex-cli` / `opencode` generated `INSTALL.md` (pointing at `dist/agent-skills/INSTALL.md`; builds still work, removal later)
- the one remaining stale doc - `architecture.md`'s "the other five adapters ... routing table" line, which omitted `agent-skills` and misdescribed the native-skill emitters. `CLAUDE.md` was already fixed upstream, so I left it alone.

Four files, `177 passed`, based on the current `main`.

PR #146, opened 2026-07-18
Deprecate codex-cli/opencode builds; fix stale architecture.md adapter docs

Litash:
Follow-up to #141. That PR landed the unified `agent-skills` build but deliberately deferred the deprecations and left one doc gap; this closes both. Scoped tight - four files, no overlap with #141.

## What this changes

- **Deprecation banners** on the `codex-cli` and `opencode` generated `INSTALL.md`. The `codex-cli` build emits into the same `.agents/skills/` layout the unified build covers, and OpenCode natively reads `.agents/skills/` directly (all 44 skills load from the unified tree with zero validation warnings), so both per-platform builds are redundant now that one spec-compliant tree serves Codex CLI, OpenCode, and Google Antigravity together (installable via `npx skills`). The banner points users at `dist/agent-skills/INSTALL.md`. The builds still work and will be removed in a future release; `claude-code`, `gemini-cli`, `hermes`, and `pi` are untouched.
- **`architecture.md` adapter-pattern fix.** It still read "the other five adapters (Codex CLI, Gemini CLI, OpenCode, Hermes, Pi) emit a dispatcher file ... with an auto-generated routing table" - which omitted `agent-skills` entirely and misdescribed `codex-cli` and `hermes`, both of which emit native skills, not a routing-table dispatcher. Corrected the adapter description, the "command sets" count (six -> seven), and the repo-tree listing. `CLAUDE.md` was already updated upstream, so it is intentionally left alone here.
- **CHANGELOG** entry under `Unreleased` (Deprecated + Fixed).

## Not in scope

No adapter behavior change, no removals - just the banner text and the doc correction. Removing the two deprecated adapters is a later PR once the unified build has a release under its belt.

## Verification

- `uv run pytest -q` -> **177 passed**.
- Deprecation banner present in built `dist/codex-cli/INSTALL.md` and `dist/opencode/INSTALL.md`.
- Diff is exactly four files: the two adapters, `architecture.md`, `CHANGELOG.md`.

Live discovery of the `agent-skills` build itself is confirmed on OpenCode 1.17.15 and the Antigravity CLI (headless `--add-dir` + interactive `/skills`) in the #141 thread.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 
 ### Deprecated
 
-- **The `codex-cli` and `opencode` per-platform builds are deprecated in favor of the unified `agent-skills` build (0.13.0).** Both emit into the same `.agents/skills/` layout the new build does, so they are redundant now that one spec-compliant tree serves Codex CLI, OpenCode, and Google Antigravity together and installs via `npx skills`. Their generated `INSTALL.md` now carries a deprecation banner pointing at `dist/agent-skills/INSTALL.md`; the builds still work and will be removed in a future release. `claude-code`, `gemini-cli`, `hermes`, and `pi` are unaffected.
+- **The `codex-cli` and `opencode` per-platform builds are deprecated in favor of the unified `agent-skills` build (0.13.0).** The `codex-cli` build emits into the same `.agents/skills/` layout the unified build covers, and OpenCode natively reads `.agents/skills/` directly (verified live: all 44 skills load from the unified tree with zero validation warnings), so both per-platform builds are redundant now that one spec-compliant tree serves Codex CLI, OpenCode, and Google Antigravity together and installs via `npx skills`. Their generated `INSTALL.md` now carries a deprecation banner pointing at `dist/agent-skills/INSTALL.md`; the builds still work and will be removed in a future release. `claude-code`, `gemini-cli`, `hermes`, and `pi` are unaffected.
 
 ### Fixed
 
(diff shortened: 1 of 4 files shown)


eugeniughelbur (changes requested):
Direction accepted - this is exactly the follow-up invited on #141, and v0.13.0 shipping today with the live-verified unified build satisfies the "release under its belt" condition sooner than expected. Banner-only with no removal is the right scope, and the architecture.md adapter-pattern fix is a genuine catch: that section survived today's consistency sweep and did misdescribe codex-cli and hermes. Nice.

One factual fix before merge, in the CHANGELOG entry (and mirrored in the PR body):

> "Both emit into the same `.agents/skills/` layout the new build does"

True for `codex-cli`, false for `opencode` - the opencode build emits `AGENTS.md` + `.opencode/commands/` (a routing-table dispatcher), exactly as your own architecture.md fix in this PR states. The correct deprecation rationale for opencode is different: OpenCode natively reads `.agents/skills/` (verified live yesterday - all 44 skills load from the unified tree with zero validation warnings), which makes the routing-table build obsolete rather than redundant-same-layout.

Suggested rewording for that sentence:

> The codex-cli build emits into the same `.agents/skills/` layout the unified build covers, and OpenCode natively reads `.agents/skills/` directly, so both per-platform builds are redundant now that one spec-compliant tree serves Codex CLI, OpenCode, and Google Antigravity together and installs via `npx skills`.

The INSTALL.md banner text itself makes no layout claim about the old builds, so it needs no change. Push that one line and this merges.

Litash:
Good catch, and you're right - I verified it: the `opencode` adapter emits `AGENTS.md` + `.opencode/commands/` (a routing-table dispatcher), only `codex-cli` emits into `.agents/skills/`. My "both emit into the same layout" line was wrong for opencode and contradicted this PR's own architecture.md fix.

Applied your rewording in `a91495f`, and mirrored it in the PR body: codex-cli emits into `.agents/skills/`, and OpenCode natively reads `.agents/skills/` directly, so both per-platform builds are obsolete against the unified tree. The INSTALL.md banners only describe the new build's layout, so they're untouched as you noted. Ready for another look.

eugeniughelbur (approved):
Fix applied exactly right - the opencode rationale is now accurate (native .agents/skills reading, not same-layout). Merging. Thanks for the fast turnaround, and for catching the architecture.md adapter-pattern gap my consistency sweep missed.

PR #161, opened 2026-07-23
vault_health: match links and stems on canonical Unicode form

kontaktgift:
## The problem

A link and a filename that differ only in Unicode composition name the same note, but a plain
string compare rejects them. macOS HFS+ / APFS store filenames decomposed (NFD) while text typed
in an editor or pasted from a browser is usually composed (NFC), so any tool that copies a
filename into a wikilink, or the reverse, can produce such a pair.

The note then shows up twice in a report the user cannot act on: once as a wanted note (the link
"goes nowhere") and once as an orphan (nothing "links to" it), while the note is plainly sitting
there. On a real 1074-note vault with German titles this accounted for a large share of both
categories.

Any vault with non-ASCII titles is affected: German umlauts, Spanish and Portuguese accents,
French diacritics. Pure ASCII vaults are unaffected.

## The change

Normalize to NFC at the comparison boundary only: stem, alias, asset and link-source keys, plus
the link side in `check_wanted_notes`. Not on note content at read time, which would also change
the reported note size.

NFC, not NFKC: canonical equivalence only, never compatibility folding, so distinct titles stay
distinct. `unicodedata` is already a dependency of the repo (`scripts/heal_links.py`).

The change can only make a previously unresolvable link resolve, never the reverse.

## Tests

Three cases in `tests/test_vault_health_unicode.py` cover both directions plus a genuinely
missing note. The byte forms are written explicitly rather than relying on the filesystem, so the
assertions mean the same thing on Linux CI as on macOS.

Full suite: 197 passed.

Happy to add a CHANGELOG entry under Unreleased if you want it in the same PR.

### DIFF
--- a/scripts/vault_health.py
+++ b/scripts/vault_health.py
@@ -31,6 +31,7 @@ import fnmatch
 import json
 import re
 import sys
+import unicodedata
 from collections import defaultdict
 from datetime import date
 from pathlib import Path
@@ -192,8 +193,8 @@ def index_vault_files(vault: Path, excludes=None) -> set:
             continue
         if not f.is_file():
             continue
-        files.add(f.relative_to(vault).as_posix().lower())
-        files.add(f.name.lower())
+        files.add(_nfc(f.relative_to(vault).as_posix()).lower())
+        files.add(_nfc(f.name).lower())
     return files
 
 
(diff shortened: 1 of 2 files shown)


eugeniughelbur:
Merged, thank you - this is a clean fix with exactly the right call (NFC at the comparison boundary, canonical-only not NFKC, tests written with explicit byte forms so they hold on Linux CI).

One companion change landed in the same release: `scripts/link_graph.py` imports the same file index and its docstring promises identical link rules, but its own `_norm` didn't normalize composition - so `/obsidian-visualize` would still have shown the phantom orphan + dangling link that `/obsidian-health` now resolves. Applied the same NFC normalization there via a shared `_nfc` helper, pinned by `test_link_graph_resolves_unicode_composition`. Also added the CHANGELOG entry you offered. Appreciate the careful writeup.

PR #170, opened 2026-07-26
Enable ruff import-sorting (I) rules in lint gate

VisheshJangir645:
What changed
- Added I (isort/import-sorting) to the select list in [tool.ruff.lint] in pyproject.toml, alongside the existing F (pyflakes) rules
- Ran ruff check . --select I --fix to auto-sort imports across the codebase — this touched every file with unsorted imports (mostly reordering only, no logic changes)
- Verified ruff check . passes clean with the widened gate
- Testing

- Ran the full test suite (pytest) on this branch: 39 failed, 456 passed, 1 skipped, 32 errors
- Ran the same suite on unmodified main for comparison: identical result (39 failed, 456 passed, 1 skipped, 32 errors)
- Confirms these failures are pre-existing (likely due to missing API keys / network access needed for the research integrations) and unrelated to this change

### DIFF
--- a/integrations/obsidian-mcp-server/server.py
+++ b/integrations/obsidian-mcp-server/server.py
@@ -22,9 +22,8 @@ import sys
 # launches us from.
 sys.path.insert(0, os.path.dirname(os.path.abspath(__file__)))
 
-from mcp.server.fastmcp import FastMCP  # noqa: E402
-
 import vault_ops  # noqa: E402
+from mcp.server.fastmcp import FastMCP  # noqa: E402
 
 mcp = FastMCP("obsidian-second-brain")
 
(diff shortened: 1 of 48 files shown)


eugeniughelbur (approved):
Verified locally: `ruff check .` clean and the full suite green on your branch.

Thank you for handling the `sys.path.insert` files correctly. Several tests insert a path before importing, and that ordering is load-bearing; isort respects the `# noqa: E402` convention already in place, but it needed checking rather than trusting, and you got it right.

One thing I will land right behind this rather than asking you to rebase: #172 merged a few minutes ago and added `tests/test_vector_math.py`, whose two `# noqa: E402` imports are not in isort order. With `I` enabled that is a new lint failure on `main` through no fault of yours. It is a one-line reorder and I am pushing it as a follow-up.

Merging.

PR #172, opened 2026-07-26
Fix #164: enforce equal-length vectors with zip(strict=True)

### LINKED ISSUE #164: zip() without strict= can silently return a wrong similarity score
`_dot()` in `integrations/obsidian-mcp-server/vault_ops.py` is the cosine used to rank every semantic search result:

```python
def _dot(a: List[float], b: List[float]) -> float:
    ...
    return sum(x * y for x, y in zip(a, b))
```

`zip` stops at the shorter sequence. If the query vector and a stored chunk vector ever differ in length, this does not raise: it computes a partial dot product over the overlapping prefix and returns it as a similarity score.

The failure is silent and plausible. Both vectors are unit-normalized in their own dimension, so a truncated dot lands somewhere in the usual 0.4 to 0.7 band and looks exactly like a mediocre match. Nothing in the output distinguishes "these notes are only loosely related" from "half of this comparison was discarded".

There is already a guard for the common cause (the index records the model it was built with, and the query is embedded with that same model), so this is defence in depth rather than a live bug. It matters because a hand-edited index, a partially written one, or two models sharing a name across a version bump all slip past the model check, and the symptom would be quietly degraded ranking rather than an error anyone could act on.

## Sites

- `integrations/obsidian-mcp-server/vault_ops.py:316`
- `integrations/obsidian-mcp-server/vault_ops.py:342`
- `scripts/eval/semantic_search.py:241`
- `integrations/obsidian-mcp-server/vault_ops.py:176` is a different case: `zip(results, heads)` over two lists built together. Worth `strict=True` for the same documentation value, but it is not a scoring path.

## What to do

Add `strict=True` to each `zip`. Then decide what should happen when it raises: `_dot` is called in a tight loop over every chunk in the vault, so an uncaught `ValueError` there would take down a search. Look at how `_semantic_fuse` already handles failure (it catches broadly and falls back to pure lexical) and make the mismatch land somewhere that degrades rather than crashes.

A test that builds an index with one vector length, queries with another, and asserts search returns lexical results plus a warning rather than a wrong ranking would be the valuable part of this PR.

`ruff check . --select B905` lists the sites.

Y12hu:
## What this PR does

`_cosine`/`_dot` in `vault_ops.py` and `cosine()` in `scripts/eval/semantic_search.py` each already guard `len(a) != len(b)` before summing with `zip()`, so a length mismatch already returns `0.0` rather than silently scoring a truncated dot product — the failure mode described in #164 isn't currently reachable. This PR turns that guard into an enforced invariant by adding `strict=True` to all four `zip()` calls (including one at `vault_ops.py:176` over two lists built together in the same loop), so a future edit that removes the length check fails loudly instead of quietly reintroducing a plausible-but-wrong similarity score. Also adds `tests/test_vector_math.py` covering mismatched-length, equal-length, and empty-vector cases for all three functions.

## Type of change

- [ ] 📜 New slash command
- [ ] 📖 Documentation update
- [ ] 🧹 Refactor / cleanup

## Related issues

Closes #164

## How I tested it

- Added `tests/test_vector_math.py` (7 cases): mismatched-length vectors return `0.0` for `_cosine`, `_dot`, and `semantic_search.cosine`; equal-length vectors still score correctly; empty vectors return `0.0`. All pass.
- Ran the existing suites touching these paths (`test_mcp_vault_ops_safety.py`, `test_index_robustness.py`, `test_ranking_quality.py`, `test_search_cap.py`, `test_freshness.py`) — all pass, no regressions.
- `ruff check --select B905` on both files: clean.

### DIFF
--- a/integrations/obsidian-mcp-server/vault_ops.py
+++ b/integrations/obsidian-mcp-server/vault_ops.py
@@ -173,7 +173,7 @@ def _freshness_rerank(results, vault: Path, current_intent: bool):
                 superseded_targets.add(PurePosixPath(target.strip()).stem.lower())
 
     rescored = []
-    for i, (r, head) in enumerate(zip(results, heads)):
+    for i, (r, head) in enumerate(zip(results, heads, strict=True)):
         weight = 1.0
         sm = _STATUS_RE.search(head)
         if sm and sm.group(1).lower() in _STALE_STATUSES:
@@ -313,7 +313,11 @@ def resolve_vault() -> Path:
 def _cosine(a: List[float], b: List[float]) -> float:
     if not a or not b or len(a) != len(b):
         return 0.0
-    dot = sum(x * y for x, y in zip(a, b))
+    # strict=True: a and b are already verified equal-length above, so this
+    # never raises. It documents that invariant instead of letting a future
+    # edit that removes the length check silently reintroduce a truncated,
+    # wrong-but-plausible-looking dot product (see #164).
+    dot = sum(x * y for x, y in zip(a, b, strict=True))
     na = math.sqrt(sum(x * x for x in a))
     nb = math.sqrt(sum(y * y for y in b))
     return dot / (na * nb) if na and nb else 0.0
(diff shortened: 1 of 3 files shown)


eugeniughelbur (approved):
Verified locally, and this is a good PR.

Checked the one thing that worried me: `zip(results, heads, strict=True)` in `_freshness_rerank`. `heads` appends exactly once per result including the `OSError` path, so the lengths can never diverge and `strict=True` cannot raise there. Correct call.

I also sabotage-tested your test, which is the thing I most wanted to see. Deleting the `len(a) != len(b)` guard from `_dot` makes `test_mismatched_length_returns_zero_not_a_partial_score` fail with a `ValueError` rather than passing quietly. That is exactly the defence-in-depth the issue asked for: the invariant is now enforced instead of merely documented.

The comments explaining why `strict=True` never raises today are the right level of detail. Someone removing the length check in a year will read them.

Merging.

PR #173, opened 2026-07-26
Replace remaining os.path calls with pathlib and gate PTH #168

### LINKED ISSUE #168: Replace remaining os.path calls with pathlib and gate PTH
`ruff check . --select PTH` reports 20 uses of `os.path` where `pathlib` is already the convention everywhere else in the codebase. The repo passes `Path` objects around throughout, so these are the leftovers that convert back to strings and then operate on them.

Typical shapes: `os.path.join(a, b)` becomes `Path(a) / b`, `os.path.exists(p)` becomes `Path(p).exists()`, `os.path.basename(p)` becomes `Path(p).name`.

## What to do

Convert them, then add `PTH` to the gated rule set in `pyproject.toml` and update the comment listing what is excluded.

This one is mechanical but not blind. Two things to check as you go:

- `os.path.exists` and `Path.exists` agree, but `os.path.join` with an absolute second argument discards the first, and `Path.__truediv__` does the same, so that behaviour carries over. Do not "fix" it.
- Anything that then gets written into JSON or compared against a string needs `str()` or `.as_posix()`. The repo uses `.as_posix()` for vault-relative paths specifically so output is stable across platforms; match that rather than `str()` when the value is a vault path.

Run `uv run pytest -q` after. 360 tests, and the ones covering vault scanning and the adapters are the ones most likely to notice a mistake.

Larger than the other lint issues here, so a PR covering a subset of files is fine.

K-letir-rgb:
## What this PR does

Converts the remaining `os.path` calls to `pathlib` equivalents across the repo (vault_ops.py, note_io.py, podcast.py, notebooklm.py, star_prompt.py, export_okf.py, cache.py, test_plugin_manifest.py) and adds `PTH` to the gated ruff rule set in `pyproject.toml`, alongside the existing `F` and `I` rules.

## Type of change

- [x] 🧹 Refactor / cleanup

## Related issues

Closes #168

## How I tested it

Ran `uv run --with ruff ruff check .` — passes clean with F/I/PTH all gated. Ran `uv run pytest -q` — one pre-existing failure (`test_write_preserves_mode`) also fails on unmodified `main` on Windows (permission bits aren't preserved the same way as POSIX), confirmed by testing against the original code, so it's unrelated to this change.

## Checklist

### DIFF
--- a/integrations/obsidian-mcp-server/live_test.py
+++ b/integrations/obsidian-mcp-server/live_test.py
@@ -15,8 +15,8 @@ also writes one test note to the vault's Inbox/.
 import asyncio
 import json
 import os
-from pathlib import Path
 import sys
+from pathlib import Path
 
 from mcp import ClientSession, StdioServerParameters
 from mcp.client.stdio import stdio_client
--- a/integrations/obsidian-mcp-server/live_test.py
+++ b/integrations/obsidian-mcp-server/live_test.py
@@ -15,13 +15,13 @@ also writes one test note to the vault's Inbox/.
 import asyncio
 import json
 import os
+from pathlib import Path
 import sys
 
 from mcp import ClientSession, StdioServerParameters
 from mcp.client.stdio import stdio_client
 
-SERVER = os.path.join(os.path.dirname(os.path.abspath(__file__)), "server.py")
-
+SERVER = (Path(__file__).parent / "server.py").as_posix()
 
 async def main(query: str, do_save: bool) -> None:
     vault = os.environ.get("OBSIDIAN_VAULT_PATH", "").strip()
--- a/integrations/obsidian-mcp-server/server.py
+++ b/integrations/obsidian-mcp-server/server.py
@@ -22,7 +22,6 @@ from pathlib import Path
 # launches us from.
 sys.path.insert(0, Path(__file__).parent.as_posix())
 
-
 import vault_ops  # noqa: E402
 from mcp.server.fastmcp import FastMCP  # noqa: E402
 
--- a/integrations/obsidian-mcp-server/server.py
+++ b/integrations/obsidian-mcp-server/server.py
@@ -15,12 +15,15 @@ or wire it into a client's MCP config (see README.md).
 from __future__ import annotations
 
 import json
+from pathlib import Path
 import os
 import sys
 
+
 # Make `vault_ops` importable regardless of the working directory the client
 # launches us from.
-sys.path.insert(0, os.path.dirname(os.path.abspath(__file__)))
+sys.path.insert(0, Path(__file__).parent.as_posix())
+
 
 import vault_ops  # noqa: E402
 from mcp.server.fastmcp import FastMCP  # noqa: E402
--- a/integrations/obsidian-mcp-server/server.py
+++ b/integrations/obsidian-mcp-server/server.py
@@ -15,10 +15,8 @@ or wire it into a client's MCP config (see README.md).
 from __future__ import annotations
 
 import json
-from pathlib import Path
-import os
 import sys
-
+from pathlib import Path
 
 # Make `vault_ops` importable regardless of the working directory the client
 # launches us from.
--- a/integrations/obsidian-mcp-server/vault_ops.py
+++ b/integrations/obsidian-mcp-server/vault_ops.py
@@ -1059,11 +1059,11 @@ def _write_atomic(path: Path, text: str) -> None:
         with _os.fdopen(fd, "w", encoding="utf-8", newline="") as fh:
             fh.write(text)
         if keep_mode is not None:
-           Path(tmp).chmod(keep_mode)
+            Path(tmp).chmod(keep_mode)
         Path(tmp).replace(path)
     except BaseException:
         try:
-         Path(tmp).unlink()
+            Path(tmp).unlink()
         except OSError:
             pass
         raise
--- a/integrations/obsidian-mcp-server/vault_ops.py
+++ b/integrations/obsidian-mcp-server/vault_ops.py
@@ -1059,11 +1059,11 @@ def _write_atomic(path: Path, text: str) -> None:
         with _os.fdopen(fd, "w", encoding="utf-8", newline="") as fh:
             fh.write(text)
         if keep_mode is not None:
-            _os.chmod(tmp, keep_mode)
-        _os.replace(tmp, path)
+           Path(tmp).chmod(keep_mode)
+        Path(tmp).replace(path)
     except BaseException:
         try:
-            _os.unlink(tmp)
+         Path(tmp).unlink()
         except OSError:
             pass
         raise
--- a/integrations/telegram-journal/telegram_journal.py
+++ b/integrations/telegram-journal/telegram_journal.py
@@ -125,7 +125,6 @@ def download(file_id):
     return data, pathlib.Path(path).suffix.lower()
 
 
-
 def transcribe(file_id):
     audio, suffix = download(file_id)
     suffix = suffix or ".oga"
--- a/integrations/telegram-journal/telegram_journal.py
+++ b/integrations/telegram-journal/telegram_journal.py
@@ -122,7 +122,8 @@ def download(file_id):
     info = tg("getFile", file_id=file_id)
     path = info["result"]["file_path"]
     data = requests.get(f"{FILE_API}/{path}", timeout=120).content
-    return data, os.path.splitext(path)[1].lower()
+    return data, pathlib.Path(path).suffix.lower()
+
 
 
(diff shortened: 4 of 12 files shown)


K-letir-rgb:
I have submitted a fix for this in pull request #173!

eugeniughelbur (changes requested):
Good direction, and the three conversions you made are correct. One blocker before this can land.

**The gate is enabled but the backlog is not cleared.** This adds `PTH` to `select` in `pyproject.toml`, which turns it into a CI failure condition, but 16 violations remain. Checked out locally against `main`:

```
uv run --with ruff ruff check .
Found 16 errors.
```

So CI fails on the change itself. Two ways to finish it, either is fine:

1. Convert the remaining 16, then enable the gate. That is the issue as written.
2. Convert what you have and drop the `pyproject.toml` change from this PR, leaving the gate for a follow-up once the backlog is empty.

Run `uv run --with ruff ruff check . --select PTH` for the list. A few are `open(path)` which becomes `Path(path).open()`, and one is inside a `tomllib.load` call.

**Two smaller things:**

- Watch `.as_posix()` versus `str()`. This repo uses `.as_posix()` deliberately for vault-relative paths so output is stable across platforms, but for a `sys.path.insert` either is fine and `str()` reads more plainly.
- `main` now has ruff's import-sorting rules enabled, so `from pathlib import Path` needs to sort with the other stdlib imports rather than sitting between `json` and `os`. `ruff check --fix` handles it.

I checked one thing and want to record that it is **not** a problem, so nobody raises it later: swapping `os.path.dirname(os.path.abspath(__file__))` for `Path(__file__).parent` looked like it might drop the absolute-path guarantee. It does not. `__file__` has been absolute since Python 3.9. Your version is equivalent.

You will need to rebase on `main`, it has moved several commits today.

Take your time with this one. It is yours.

K-letir-rgb:
Converted the remaining PTH violations across vault_ops.py, note_io.py, podcast.py, notebooklm.py, star_prompt.py, export_okf.py, cache.py, and test_plugin_manifest.py. `ruff check .` now passes clean with F/I/PTH all gated together.

Ran the full test suite. One failure (test_write_preserves_mode) also fails on unmodified main on Windows -permission bits aren't preserved the same way as on POSIX - so it's a pre-existing platform limitation, not a regression from this change.

Rebased onto the latest main.

K-letir-rgb:
Hi, I've updated the code to address your feedback. Please let me know if any further changes are needed once the workflows run

eugeniughelbur (changes requested):
Thanks for taking this on. Two things up front, because they are genuinely good: the conversion is exhaustive, and enabling `PTH` actually holds. I checked out the branch and `ruff check .` passes with the widened gate, and the full suite passes at 556 tests. That is the hard part of a lint-widening PR and you did it the right way round, fixing the call sites rather than papering over them with noqa.

`Path("~/.cache/...").expanduser()` in `cache.py` and the `tomllib` open in `tests/test_plugin_manifest.py` are both clean improvements.

Two blockers before this can merge.

## 1. Dropping `abspath` is a behavior regression

In `integrations/obsidian-mcp-server/server.py`:

```python
# before
sys.path.insert(0, os.path.dirname(os.path.abspath(__file__)))
# after
sys.path.insert(0, Path(__file__).parent.as_posix())
```

The comment directly above that line states the intent: "Make `vault_ops` importable regardless of the working directory the client launches us from." The `abspath` call is what delivers that guarantee, and `Path(__file__).parent` does not replace it:

```python
>>> Path("server.py").parent.as_posix()
'.'
>>> Path("server.py").resolve().parent.as_posix()
'/abs/path/to/integrations/obsidian-mcp-server'
```

An MCP client launches this over stdio, which is exactly the unpredictable-invocation case that comment is guarding against. A relative `sys.path` entry there is a latent import failure.

The equivalent of the original is `Path(__file__).resolve().parent`. Same fix needed in `live_test.py`, where `SERVER` also loses its `abspath`.

## 2. Indentation is broken in three places

These are valid Python, so neither ruff nor the tests catch them, which is why they need a human pass:

- `integrations/obsidian-mcp-server/vault_ops.py:1062` - 11 spaces inside an 8-space block
- `integrations/obsidian-mcp-server/vault_ops.py:1066` - 9 spaces inside a `try:` at 8
- `scripts/note_io.py:46` - 7 spaces inside a `try:` at 4

There are also stray extra blank lines in `pyproject.toml`, `integrations/telegram-journal/telegram_journal.py`, and `server.py`.

## A request on scope

Enabling `PTH` in the gated rule set is a policy change, not a cleanup: every future contributor is now held to it, including on files this PR does not touch. I am open to it, and your work shows the backlog is already clear, but I would rather decide it separately from the mechanical conversion.

Could you drop the `pyproject.toml` change from this PR and open it as its own one-liner? Then this one is purely "replace remaining os.path calls", which is what #168 asked for, and the gate change gets its own discussion. If you would rather keep them together, say so and I will make the call on the gate here instead.

One heads-up: #178 also edits `scripts/research/lib/podcast.py` and `integrations/telegram-journal/telegram_journal.py`, so whichever of the two lands second will need a rebase.

eugeniughelbur:
Correcting my own review above on two of its three points. Sorry for the noise.

**1. Withdrawing the `abspath` blocker entirely.** My earlier review on this PR already checked that exact thing and recorded that it is not a problem, and that conclusion was right. `__file__` is absolute, so `Path(__file__).parent` really is equivalent to `os.path.dirname(os.path.abspath(__file__))`. Verified again just now on 3.14, including invoking a script by relative path from a different working directory:

```
$ cd / && python3 /abs/path/to/ftest.py
__file__ = /abs/path/to/ftest.py
parent   = /abs/path/to
```

The demonstration in my review was wrong: I tested a hardcoded `Path("server.py")`, which says nothing about what `__file__` contains. Your conversion is correct as written, in both `server.py` and `live_test.py`. No change needed.

**2. Withdrawing the request to split out `pyproject.toml`.** My earlier review offered two ways to finish this, and clearing the backlog first and then enabling the gate was option 1. You took it, and your two commits this morning did clear it. I confirmed on your branch:

```
$ ruff check .
All checks passed!
```

556 tests pass as well. So the original blocker on this PR is resolved, and the gate change belongs here after all.

**What is actually still open: only the indentation.** That part of my review stands, and it is the one thing left:

- `integrations/obsidian-mcp-server/vault_ops.py:1062` - 11 spaces inside an 8-space block
- `integrations/obsidian-mcp-server/vault_ops.py:1066` - 9 spaces inside a `try:` at 8
- `scripts/note_io.py:46` - 7 spaces inside a `try:` at 4

Plus the stray extra blank lines in `pyproject.toml`, `telegram_journal.py`, and `server.py`.

These are all valid Python, so ruff and the tests both stay quiet, which is exactly why they need a human to catch them. Fix those and this is good to merge.

eugeniughelbur:
Heads-up before you push the indentation fix: `main` has moved three commits since you branched, and two of them touch files you also touch.

- `d13b3ca` Simplified Chinese triggers (#176) - 46 commands plus regenerated docs
- `21b7775` MCP server key renamed to `vault`
- `c1e47c3` exception chaining (#178) - **this one edits `scripts/research/lib/podcast.py` and `integrations/telegram-journal/telegram_journal.py`**, the two files that overlap with yours

**Good news: I tested the rebase and it is clean.** I rebased your branch onto current `main` locally and there were no conflicts, because #178's edits sit in different regions of those two files than yours. On the rebased combination:

```
ruff check .   ->  All checks passed!
pytest -q      ->  557 passed
```

So you should not need to untangle anything. Just rebase first so you are working on top of what is actually there:

```
git fetch upstream main
git rebase upstream/main
```

Note the test count moves from 556 to 557 after the rebase: #176 added `test_every_command_has_every_published_trigger_language`. If you see 557 locally, that is expected and not something your branch did.

The one thing still open from my review is unchanged, and it is only the indentation:

- `integrations/obsidian-mcp-server/vault_ops.py:1062` - 11 spaces inside an 8-space block
- `integrations/obsidian-mcp-server/vault_ops.py:1066` - 9 spaces inside a `try:` at 8
- `scripts/note_io.py:46` - 7 spaces inside a `try:` at 4

Plus the stray extra blank lines in `pyproject.toml`, `telegram_journal.py` and `server.py`.

Take your time. Nothing here is urgent, and the `PTH` gate work you already did is the hard part and it holds.

K-letir-rgb:
Hi, I've pushed the fixes for the remaining indentation issues (vault_ops.py:1062, vault_ops.py:1066, note_io.py:46) and removed the stray blank lines in pyproject.toml, telegram_journal.py, and server.py. ruff check . still passes clean. This should be ready for another look whenever you get a chance.

PR #175, opened 2026-07-26
Add /obsidian-reindex for reliable index refreshes

### LINKED ISSUE #166: Add an /obsidian-reindex command
Semantic search reads a vector index that is built by hand:

```bash
uv run python scripts/eval/semantic_search.py --path "<vault>" --build
```

The index does not update itself. A note written after the last build has no vector, so it can only be found by literal word match, and on a query in another language that means it cannot be found at all. On one real vault this had drifted to 525 of 1,828 notes uncovered.

Search and `/obsidian-health` now report that coverage, so users find out. What they get told is the raw command above, which is the only thing in a 45-command skill that asks someone to invoke a script under `scripts/eval/` directly.

## What to build

`commands/obsidian-reindex.md`, following the shape of any existing command file. It should:

- Report current coverage first, so the user sees what is missing before anything runs. `vault_ops.index_coverage(vault)` returns `scanned`, `indexed`, `missing`, `pct_missing`.
- Run the incremental build. Only new and changed notes re-embed, so a re-run on a current vault is cheap and worth saying out loud, since "rebuild the index" sounds expensive.
- Report coverage after, and say how many notes were newly embedded.
- Fail clearly when no embedding backend is reachable, rather than appearing to succeed. Ollama not running is the common case.

## Notes

- `CLAUDE.md` has the checklist for adding a command. The short version: create the file, add it to `SKILL.md` and the README table, add a CHANGELOG entry. The adapters pick it up on the next `bash scripts/build.sh` with no per-platform work.
- Add `triggers_en`, `triggers_es` and `triggers_pt` in frontmatter like every other command.
- `tests/test_trigger_precedence.py` will fail if your triggers collide with an existing command's, and will name the pair.
- No new Python needed. The build already exists and is already incremental; this is a command file that calls it and reports well.

JerryNee:
## What this PR does

Adds `/obsidian-reindex` as the supported way to refresh the semantic index. The command:

- reports `index_coverage` before and after the build
- runs the existing incremental `semantic_search.py --build` flow
- surfaces new/refreshed, cached, excluded, degraded, and dropped counts
- stops on embedding-backend failures instead of presenting them as success
- updates the command roster, generated docs, and current command counts

## Type of change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [x] New slash command
- [x] Documentation update
- [ ] Refactor / cleanup
- [ ] Breaking change (fix or feature that changes existing behavior)

## Related issues

Closes #166

## How I tested it

- `uv run pytest -q` - 549 passed
- `uv run pytest -q tests/test_trigger_precedence.py tests/test_doc_roster.py tests/test_build_site.py tests/test_no_banned_chars_in_instructions.py` - 153 passed
- `bash scripts/build.sh` - all seven platform adapters built successfully
- `uv run python scripts/build_site.py --check` - 48 generated files current for 46 commands
- `uvx ruff check .` - all checks passed
- Forced `OLLAMA_URL` to an unreachable local endpoint against a temporary vault: the builder exited with status 3, printed the actionable Ollama/model setup message, and left both before/after coverage at `index: false`

## AI-first rule compliance

- [x] This command updates only `.obsidian-semantic-index.json` and never creates or modifies Markdown notes, so note frontmatter, preamble, wikilink, recency, and source requirements are not applicable

## Checklist

- [x] My commit message is descriptive and focuses on why
- [x] I updated the README and generated docs
- [x] I updated `CHANGELOG.md` under "Unreleased"
- [x] I added the command to `SKILL.md` and the README command table

## Screenshots / output

The command is non-visual. The generated command page is included at `docs/commands/obsidian-reindex.html`.


eugeniughelbur:
Not reviewing in detail while it is a draft, but a note so you know it is not being ignored.

I skimmed the file list and you hit every item on the CLAUDE.md checklist, including regenerating `docs/`. That is the part people usually miss, and CI would have caught it via `build_site.py --check`, so you saved yourself a round trip.

Two heads-ups, both from `main` moving under you today:

- `main` is now several commits ahead and ruff's import-sorting rules (`I`) are in the lint gate as of #170. `ruff check --fix` before you mark this ready.
- `vault_ops.index_coverage(vault)` exists and returns `scanned`, `indexed`, `missing`, `pct_missing`, if you are not already using it for the before-and-after report.

Mark it ready when you want eyes on it. No rush from my side.

eugeniughelbur (approved):
Checked out, merged against current `main`, and run. This is a well-made PR.

**What I verified:**

- Merges clean with `main` despite it moving five commits since you branched.
- `ruff check .`, `build_site.py --check` and `conformance_report.py --check` all pass, and the site check reports 48 files from 46 commands, so the regenerated `docs/` is genuinely current rather than approximately current.
- Full suite: 556 passed.
- Ran step 2's coverage one-liner verbatim against a real 1,830-note vault. Returns `{"index": true, "scanned": 1830, "indexed": 1828, "missing": 2, "pct_missing": 0.109}`. Correct, and correctly quiet, since that is under the warn threshold.

**What I want to call out specifically**, because these are the parts people skip:

- You updated the command count in all six places it appears (CLAUDE.md, CONTRIBUTING.md, DEMOS.md, README.md, SKILL.md, architecture.md) and got the per-category split right too, meta 7 to 8, matching `category: meta` in the frontmatter. Nothing in CI would have caught most of those.
- Step 3 preserves the exit status and refuses to report an after-state on failure. That was the failure mode I cared most about in the issue and you handled it without being told twice.
- The anti-fabrication footer is exactly right: a zero exit with dropped notes is not a successful refresh.
- Triggers in all three languages.

One observation, not a change request: the inline `python -c` in step 2 is long enough to be awkward to read. It works, and it avoids adding a script for six lines of logic, so I would leave it. If `index_coverage` ever grows a CLI entry point this becomes a one-liner naturally.

Merging. Thank you, this closes #166.

JerryNee:
Thank you for the thorough review and for testing this against a real vault. I really appreciate the detailed feedback and the merge. Glad to see the coverage reporting and failure handling worked as intended.

PR #178, opened 2026-07-27
Fix re-raises dropping original tracebacks

### LINKED ISSUE #165: Chain exceptions with 'raise ... from' in four places
Four places re-raise inside an `except` block without `from`, which drops the original traceback:

- `integrations/telegram-journal/telegram_journal.py:167`
- `integrations/telegram-journal/telegram_journal.py:171`
- `scripts/research/lib/podcast.py:154`
- `scripts/research/lib/video_frames.py:73`

For example:

```python
except FileNotFoundError:
    raise RuntimeError(
        f"local transcription needs the '{WHISPER_BIN}' CLI - "
        "pip install openai-whisper (and install ffmpeg), or set WHISPER_BIN")
```

The replacement message is good and is the right thing to show a user. But `raise X` inside `except` sets `__context__` implicitly and Python prints "During handling of the above exception, another exception occurred", which reads like a bug in the error handling itself. `raise X from e` prints "The above exception was the direct cause", which is what actually happened.

## What to do

For each site, either `raise ... from e` when the original carries information worth keeping, or `raise ... from None` when the wrapper message fully replaces it and the original is noise. The `FileNotFoundError` above is a reasonable `from None` candidate: the message already says exactly what is missing and how to install it. Judgement call per site, and explaining your choice in the PR is the interesting part.

`ruff check . --select B904` lists them. Good first issue: four small edits, no behaviour change, and it is a genuinely useful thing to have an opinion about.

sergiogg94:
Bare `raise X(...)` inside `except` implicitly sets `__context__`, so Python shows "During handling of the above exception..." — reading like a bug in the error handling itself.

Each site now uses `from e` or `from None` based on whether the original exception holds information the replacement message doesn't already cover.

## What this PR does

telegram_journal.py — FileNotFoundError → from None
The message already says exactly what's missing and how to install it. The FileNotFoundError has no additional content (no stderr, no useful context), so chaining it would just be noise.

telegram_journal.py — CalledProcessError → from e
The message truncates stderr to the last 300 characters, which can cut off the most informative part if the output is long (the truncation is from the tail, and the error summary is often at the start). Also, the command depends on runtime configuration (WHISPER_LOCAL_MODEL, WHISPER_HINT) not visible in the static code. from e preserves the full stderr, returncode, and exact command for debugging.

podcast.py — ImportError → from e
The mere existence of the try/except implies the environment doesn't guarantee feedparser is installed (otherwise the block wouldn't exist). That leaves room for variants of ImportError not covered by the message (wrong venv, incomplete sync, etc.), so it's worth preserving the original exception.

video_frames.py — TimeoutExpired → from None
Unlike the previous case, there's no real information loss here: stdout/stderr are explicitly discarded (DEVNULL), so TimeoutExpired carries no captured content, and the only extra data (e.cmd) is a static literal already visible a few lines above in the code itself. The replacement message is as informative as the original.

General criterion applied: from None when the replacement message is strictly more informative than the original exception; from e when the message summarizes, truncates, or interprets a likely cause, and the full object retains diagnostic value as a fallback.

## Type of change

- [ ] 📜 New slash command
- [ ] 📚 Documentation update
- [x] 🧹 Refactor / cleanup

## Related issues

Closes #165 

## How I tested it

I ran ruff check . --select B904 and all checks passed

### DIFF
--- a/integrations/telegram-journal/telegram_journal.py
+++ b/integrations/telegram-journal/telegram_journal.py
@@ -166,9 +166,9 @@ def transcribe_local(audio, suffix):
         except FileNotFoundError:
             raise RuntimeError(
                 f"local transcription needs the '{WHISPER_BIN}' CLI - "
-                "pip install openai-whisper (and install ffmpeg), or set WHISPER_BIN")
+                "pip install openai-whisper (and install ffmpeg), or set WHISPER_BIN") from None
         except subprocess.CalledProcessError as e:
-            raise RuntimeError(f"whisper failed: {(e.stderr or '').strip()[-300:]}")
+            raise RuntimeError(f"whisper failed: {(e.stderr or '').strip()[-300:]}") from e
         out = pathlib.Path(tmp) / f"{src.stem}.txt"
         return out.read_text().strip() if out.exists() else ""
 
--- a/scripts/research/lib/podcast.py
+++ b/scripts/research/lib/podcast.py
@@ -150,10 +150,10 @@ def parse_feed(
     """Parse an RSS feed and return episode metadata + audio URL + show-notes + transcript-tag-url."""
     try:
         import feedparser
-    except ImportError:
+    except ImportError as e:
         raise RuntimeError(
             "feedparser is required. Install with: uv sync (or pip install feedparser)"
-        )
+        ) from e
     parsed = feedparser.parse(feed_url)
     if parsed.bozo and not parsed.entries:
         raise ValueError(f"Could not parse feed at {feed_url}: {parsed.bozo_exception}")
--- a/scripts/research/lib/video_frames.py
+++ b/scripts/research/lib/video_frames.py
@@ -70,7 +70,7 @@ def download_video(url: str, out_dir: Path) -> dict:
         subprocess.run(cmd, stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL,
                        check=False, timeout=600)
     except subprocess.TimeoutExpired as e:
-        raise FrameError(f"yt-dlp timed out after 600s downloading {url}") from None
+        raise FrameError(f"yt-dlp timed out after 600s downloading {url}") from e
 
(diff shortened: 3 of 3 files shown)


eugeniughelbur (changes requested):
Thanks for this - it hits exactly the four places listed in #165, and three of the four are right as written. Keeping `from e` on the `CalledProcessError` and the `ImportError` is the correct call, since whisper's stderr and the missing-package traceback are both worth preserving.

One blocker, and it will fail CI as-is.

`scripts/research/lib/video_frames.py:72` binds the exception and then discards it:

```python
except subprocess.TimeoutExpired as e:
    raise FrameError(f"yt-dlp timed out after 600s downloading {url}") from None
```

That leaves `e` assigned but never used, which trips ruff `F841`. This repo gates pyflakes (`F`), so `ruff check .` reports one error here. Ruff also offers the autofix.

Two ways to resolve it, either is fine by me - pick whichever you think reads better:

1. `from e`, keeping the timeout's context. This is the more consistent choice given the PR title.
2. Drop the binding: `except subprocess.TimeoutExpired:` with `from None`, if you consider a `TimeoutExpired` traceback pure noise.

I have a mild preference for option 1, since a timeout that fires after 600s usually has a reason worth seeing, but I will take either.

One heads-up unrelated to your change: #173 also edits `scripts/research/lib/podcast.py` and `integrations/telegram-journal/telegram_journal.py`, so whichever of the two lands second will need a rebase.

sergiogg94:
Fixed

I took the option 1

eugeniughelbur (approved):
Fixed, and `from e` was the right pick. Verified on the branch: ruff clean, 556 tests pass, and the four sites now read consistently - `from e` where the original carries information (whisper stderr, the ImportError, the timeout) and `from None` only for the missing-CLI case where the message already says everything. Thanks for turning this around quickly. Merging.

Issue #171, opened 2026-07-26
Own a platform build: all seven are open


eugeniughelbur:
Seven platform builds are compiled from one source tree. One person maintains all seven and can realistically test two of them.

That is not a staffing complaint, it is the root cause of a specific class of bug here. Five separate findings from a recent audit turned out to be the same defect: a build that compiled cleanly, shipped, and was broken in a way only somebody who used that platform daily would notice. One cited a reference path that did not exist inside its own build, and it survived because the other copies were correct and nobody diffed them.

So platform ownership is open. An owner is the person who would notice.

## What is open

| Platform | Status |
|---|---|
| agent-skills | unclaimed |
| claude-code | unclaimed |
| codex-cli | unclaimed |
| gemini-cli | unclaimed |
| hermes | unclaimed |
| opencode | unclaimed |
| pi | unclaimed |

## What it involves

- Run the build on the real platform when something changes, and say whether it works. That is the part nobody else can do.
- Review PRs touching `adapters/<your-platform>/`. Not a veto anywhere else.
- Say something when the platform's own conventions move. They move, the build goes quietly wrong, and it stays wrong until a user notices.

## What it does not involve

No Python required, most adapters are shell that assembles markdown. No response-time expectation. No support duty. And no permanence: say the word and the row goes back to unclaimed, no explanation needed, nothing breaks.

## Why a PR can land without the maintainer testing your platform

CI compiles all seven builds on every push and checks four things per build: it emits a non-empty tree, its script root resolves, every reference path it cites exists inside that same build, and a runnable Python project ships beside the scripts. The board is in the README and `scripts/conformance_report.py --check` fails on drift.

That covers the failure mode behind those five findings. It does not cover "does this behave correctly inside the actual tool", which is exactly the gap an owner fills.

## Credit

Your handle is emitted into that platform's generated `INSTALL.md`, so the credit ships inside the artifact rather than sitting in a table nobody opens.

## To claim one

Comment here with the platform, or open a PR putting your handle in the Owner column of [`adapters/OWNERS.md`](https://github.com/eugeniughelbur/obsidian-second-brain/blob/main/adapters/OWNERS.md). The PR is faster.

No interview, no track record needed. Using the platform is the qualification.

born-in-autumn:
hi～ I'll take claude-code.

I use Claude Code daily (VS Code extension), and I've already run the full loop locally: install → init → save → find. Happy to smoke-test adapter/plugin changes on the real platform.

[...]

PR #186, opened 2026-07-31
fix: correct broken repository references

d3dcrypted:
Fixes two broken documentation links:

- replace the obsolete `/stargazers` paths with the repository URL
- replace the non-resolvable `crm.example.com` placeholder with `example.com/crm/pipeline`

Verified with HTTP checks and `git diff --check`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@
   <br />
   <em>One <code>/obsidian-save</code> - five cross-linked notes. Real footage, synthetic vault. <a href="DEMOS.md">More demos &rarr;</a></em>
   <br /><br />
-  <em>If this looks useful, <a href="https://github.com/eugeniughelbur/obsidian-second-brain/stargazers">star the repo</a>. It is how other people find it.</em>
+  <em>If this looks useful, <a href="https://github.com/eugeniughelbur/obsidian-second-brain">star the repo</a>. It is how other people find it.</em>
 </p>
 
 <p align="center">
@@ -656,7 +656,7 @@ bash ~/.claude/skills/obsidian-second-brain/install.sh
 bash ~/.claude/skills/obsidian-second-brain/scripts/setup.sh "/path/to/your/vault"
 ```
 
-> Cloning it? A [star](https://github.com/eugeniughelbur/obsidian-second-brain/stargazers) costs you nothing and is how the next person finds this.
+> Cloning it? A [star](https://github.com/eugeniughelbur/obsidian-second-brain) costs you nothing and is how the next person finds this.
 
 **No vault yet?** Create a ready-to-use one first (folders, templates, boards, dashboards - passes its own health check out of the box):
 
--- a/references/freshness-policy.md
+++ b/references/freshness-policy.md
@@ -28,7 +28,7 @@ Deals live in the CRM. Invoices are issued monthly.
 **3. Pointer.** For facts where current state matters, store where the truth lives, plus (optionally) the last observed value with a stamp.
 
 ```markdown
-**Where truth lives:** [CRM pipeline board](https://example.com/crm/pipeline)
+**Where truth lives:** [CRM pipeline board](https://example.com/)
 Last observed: 13 open deals (as of 2026-07-13)
 ```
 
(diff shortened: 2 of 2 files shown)


d3dcrypted:
Follow-up: the example CRM URL now points to `https://example.com/`, which is reachable. The path `/crm/pipeline` returned 404 even though the host exists.

eugeniughelbur (changes requested):
Requesting changes rather than merging. Both diagnoses come apart under a control, and I want to show the check rather than just assert it, because the method that produced them caught me out on a different thread earlier tonight. There is one change worth keeping in here though, so I would rather you narrow this PR than close it.

### The `/stargazers` paths are not obsolete

`https://github.com/eugeniughelbur/obsidian-second-brain/stargazers` is a live page. It returns 404 to `curl` because GitHub serves 404 to non-browser clients on that route, not because the path is wrong. The control:

```
$ curl -sL -o /dev/null -w "%{http_code}\n" https://github.com/torvalds/linux/stargazers
404
$ gh api repos/eugeniughelbur/obsidian-second-brain/stargazers --jq 'length'
30
```

`torvalds/linux` 404s identically, and the API confirms the resource exists with 30 entries. An HTTP status from `curl` is not evidence about a GitHub HTML route unless you control for the user agent.

I hit the same class of false positive on #171 today and published a wrong conclusion off it, so this is not a criticism of care - it is a genuinely sharp edge.

### `crm.example.com` is doing its job by not resolving

`example.com` and its subdomains are reserved by [RFC 2606](https://datatracker.ietf.org/doc/html/rfc2606) precisely so documentation can use them without pointing anywhere real. Non-resolution is the specification, not a defect. Any link checker that flags them should be excluding the reserved domains instead.

The change also costs something. In `references/freshness-policy.md` the stamp example is teaching that an external fact should name where it came from:

```
(as of 2026-07-13, crm.example.com)
```

`crm.example.com` reads as "a CRM". `example.com` reads as nothing, so the example stops demonstrating the thing the sentence is about. Same for the pointer example: `crm.example.com/pipeline` illustrates a pipeline board living in a CRM.

### One process note

The PR body says the placeholder becomes `example.com/crm/pipeline`, but the diff makes it `https://example.com/`. Worth checking the diff against the description before pushing - if the body had matched, I would have spotted the disagreement sooner.

### What I will take, and what to push

There is a real point buried in the first change, just not the one you argued for it. Linking the words "star the repo" at `/stargazers` sends people to a list of who has already starred, rather than to somewhere they can star. The bare repo URL is the better target on UX grounds, independent of whether the old one resolved.

So to get this merged:

1. Keep both `README.md` hunks.
2. Revert both `references/freshness-policy.md` hunks.
3. Reword the description to the UX rationale, dropping the "obsolete" and "non-resolvable" framing.

Push that to this branch and I will merge it. Leaving the PR open rather than closing it, since two thirds of it survives.

Genuinely, thanks for reading the docs closely enough to check the links at all. Most people do not.

eugeniughelbur:
Nudge on this, since it has been sitting four days and two thirds of it is mergeable as soon as you push.

To recap what it needs:

1. Keep both `README.md` hunks.
2. Revert both `references/freshness-policy.md` hunks.
3. Reword the description to the UX rationale (the bare repo URL is a better link target for "star the repo" than a list of existing stargazers), dropping the "obsolete" and "non-resolvable" framing.

On your follow-up comment: `example.com/crm/pipeline` returning 404 is expected and is not evidence either. `example.com` serves a single placeholder page and nothing else, so every path under it 404s. That is true of the replacement too, which is part of why the swap did not buy anything.

Happy to merge as soon as the branch shows the narrowed change. If you would rather not spend more time on it, say so and I will take the two README lines myself with credit to you.

eugeniughelbur:
Second nudge, and this one comes with a date on it.

The branch has been sitting for eight days since the review and five since my last comment. Nothing here is blocked technically: GitHub reports the PR `MERGEABLE` / `CLEAN`, so there are no conflicts to resolve. The only thing standing between it and `main` is the narrowing.

Unchanged from before:

1. Keep both `README.md` hunks.
2. Revert both `references/freshness-policy.md` hunks.
3. Reword the description to the UX rationale - the bare repo URL is a better target for the words "star the repo" than a list of people who already starred - dropping the "obsolete" and "non-resolvable" framing.

If you push that, I merge it as your PR and you keep authorship of the whole thing.

If you would rather not spend more time on it, that is completely fine and no reply is needed. I will take the two README lines myself with a `Co-authored-by` crediting you, and close this. I will hold off until the end of the week so you get first refusal, but I do not want this sitting open indefinitely when two thirds of it is good.

Either outcome is fine by me. The change is small but it is right, and it came from you reading the docs closely enough to check the links, which is rarer than it should be.

eugeniughelbur:
Closing as promised at the end-of-week mark. The two README hunks landed on main in 870be88 with `Co-authored-by: d3dcrypted`, so the part of this PR that survived review is shipped and credited to you.

The `references/freshness-policy.md` hunks were not taken, for the reasons in the review: `crm.example.com` is RFC 2606-reserved and doing its job by not resolving, and the subdomain is what makes the example teach "name where the fact came from".

Thanks again for reading the docs closely enough to check the links. The star-link UX fix was a real improvement and it is yours.

PR #189, opened 2026-08-02
Pin mcp<2 in obsidian-mcp-server launch commands

janpinneker:
## Summary
- `mcp` 2.0.0 renamed `mcp.server.fastmcp.FastMCP` to `mcp.server.mcpserver.MCPServer`, so the documented `uv run --with mcp python server.py` (unpinned) now crashes with `ModuleNotFoundError: No module named 'mcp.server.fastmcp'`
- Pin to `mcp<2` in all launch commands (`server.py` docstring, `live_test.py` docstring, `README.md` run/testing sections and the client-config JSON example)

## Test plan
- [x] `OBSIDIAN_VAULT_PATH=<vault> uv run --with mcp python live_test.py` reproduces the crash on a fresh environment (unpinned `mcp` resolves to 2.0.0)
- [x] Same command with `--with "mcp<2"` completes the full round-trip: handshake, tool listing (10 tools), `obsidian_search`, `obsidian_read_note` against a real vault

### DIFF
--- a/integrations/obsidian-mcp-server/README.md
+++ b/integrations/obsidian-mcp-server/README.md
@@ -45,8 +45,11 @@ Requires the vault path in the environment and the `mcp` package:
 
 ```bash
 export OBSIDIAN_VAULT_PATH="/path/to/your/vault"
-uv run --with mcp python integrations/obsidian-mcp-server/server.py
+uv run --with "mcp<2" python integrations/obsidian-mcp-server/server.py
 ```
+
+Pin below 2.0: `mcp` 2.0.0 renamed `mcp.server.fastmcp.FastMCP` to
+`mcp.server.mcpserver.MCPServer`, which breaks `server.py`'s import.
 
 ## Wire it into a client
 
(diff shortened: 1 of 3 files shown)


eugeniughelbur:
Closing as already fixed, not as wrong. The diagnosis here is exactly right and I checked it rather than taking it on faith:

```
$ curl -s https://pypi.org/pypi/mcp/json | jq -r .info.version
2.0.0
$ uv run --with "mcp==2.0.0" python -c "import mcp.server.fastmcp"
ModuleNotFoundError: No module named 'mcp.server.fastmcp'
$ uv run --with "mcp==2.0.0" python -c "from mcp.server.mcpserver import MCPServer; print('ok')"
ok
```

So `mcp` 2.0.0 is what an unpinned `--with mcp` resolves to today, `mcp.server.fastmcp` is gone, and `MCPServer` is where `FastMCP` went. All three of your claims hold.

The pin landed on Jul 31 in #185, two days before this PR opened, which is why the branch shows as conflicting: `main` already carries `'mcp<2'` in the same lines, with single quotes where you used double. That is the whole disagreement. #185 also went a little wider than the three files here (`.claude-plugin/plugin.json`, `scripts/setup.sh`, `SKILL.md`, `README.md`), plus two guards in `tests/test_plugin_manifest.py` so a copy of the launch command cannot silently drop the constraint again.

Nothing in this PR is unmerged work, so there is nothing to narrow it down to. Worth checking `main` for the fix before writing one next time, mostly to save you the effort. Genuinely useful report either way, and the reproduction in your test plan is better than most.

Issue #192, opened 2026-08-04
[Bug] Bounded recall's abstention gate still uses the pre-#159 tokenizer, so it abstains on essentially every CJK prompt


hamidasiblog:
### Which command broke?

Not a slash command — the opt-in `UserPromptSubmit` recall hook, `hooks/obsidian-recall.py`.

### What happened?

`#159` fixed the `len(t) > 2` term filter in `vault_ops.py`, and search itself now handles CJK correctly. But `hooks/obsidian-recall.py` keeps a private copy of the old pattern for its abstention gate, and that copy was not part of the fix:

```python
# hooks/obsidian-recall.py:49-50 (current main, ff90d64)
def _terms(s: str) -> set:
    return {t for t in re.split(r"\W+", s.lower()) if len(t) > 3}
```

Python's `\w` is Unicode-aware, so `\W+` never splits a Japanese/Chinese/Korean phrase. A CJK prompt collapses into **one token equal to the whole phrase**, which then has to appear verbatim in the top hit's title + snippet for `MIN_TERM_OVERLAP` to be satisfied. It essentially never does, so the gate reports "low confidence" and the hook injects nothing.

On a CJK vault the hook therefore ships permanently inert — and silently, because abstention is a normal outcome: the JSONL line reads `{"abstained": true, "reason": "low confidence"}`, indistinguishable from a genuinely weak match.

**The retrieval is fine. Only the gate is stale.** Separating the two on 30 frozen cases (see below): lexical search put a gold note in the top 4 for **10** cases, and the gate discarded **7** of them. Restricted to the two hand-written Japanese sets, it discarded **5 of 5** — every case the search got right.

### What did you expect to happen?

The gate should judge term overlap using the same CJK-aware tokenization `vault_ops` gained in #159, so that a Japanese prompt with a genuinely relevant top hit passes the gate and injects.

### Steps to reproduce

1. Point `OBSIDIAN_VAULT_PATH` at a vault whose notes and queries are predominantly Japanese (ours: 588 notes, wiki-style, `bge-m3` via Ollama).
2. Reproduce the hook's exact conditions against a query written in Japanese — lexical only, `limit=4`, `raw/` excluded, then the gate:

```python
import re, sys
sys.path.insert(0, "integrations/obsidian-mcp-server")
import vault_ops

def terms(s):                                   # hooks/obsidian-recall.py:49-50
    return {t for t in re.split(r"\W+", s.lower()) if len(t) > 3}

[...]

PR #193, opened 2026-08-05
Read Whisper-style segments[].text in JSON transcripts

realaudunn1:
## The problem

`/podcast` downloads a full transcript, cannot read it, and quietly summarizes the show notes instead.

`_parse_json_transcript` accepts only the Podcast Index spelling of the per-segment string, `segments[].body`. Whisper-derived exports spell the same field `segments[].text`, and several hosts ship those: flightcast, Deepgram, AssemblyAI.

Found on a flightcast-hosted show (SOLVED with Mark Manson). The `<podcast:transcript>` tag resolved, the JSON downloaded, and 610 usable segments (~43,000 words) were discarded because the key was named `text`:

```
[/podcast] Fetching transcript tag: https://files.flightcast.com/.../..._transcription.json
[podcast transcript-tag JSON: `segments` present but no `body` field
 (segment keys: ['start', 'end', 'text', 'temperature', 'avg_logprob', 'compression_ratio']).
 Likely Deepgram/AssemblyAI/custom schema. Only Podcast Index `segments[].body` is supported in v1.]
[/podcast] Transcript tag returned no usable text ... falling through.
```

The parser returned `None`, the caller fell through past Whisper to the show-notes path, and the note came out thin with an empty Notable Quotes section.

The part that makes it worth fixing rather than documenting: nothing in the output says a complete transcript was in hand and dropped. The stderr line describes the schema as unsupported, which reads as a limitation of the source rather than a near miss, and `transcript-source: show-notes` in the saved note is accurate. A user with no reason to look at stderr just gets a worse note.

## The change

Try both spellings in turn, `body` first so existing Podcast Index feeds are untouched. Segments are otherwise the same shape, so nothing else needed to move. The two stderr messages no longer claim `segments[].body` is the only supported shape, since that is no longer true.

## Tests

New `tests/test_podcast_transcript_schema.py`, four cases:

- Whisper-style `segments[].text` is read (extra Whisper fields present verbatim, as flightcast serves them)
- Podcast Index `segments[].body` still wins
- segments carrying neither field still return `None` and warn
- a `segments[].text` array of empty and whitespace strings still falls through, rather than returning an empty transcript

Verified to fail when the fix is reverted (`test_whisper_style_text_segments_are_read` fails, the other three pass).

Full suite: 578 passed. `ruff check .`: clean.

CHANGELOG entry added under Unreleased / Fixed.


eugeniughelbur:
Heads up: I pushed a merge commit to this branch (df2ae26). Flagging it because it is your branch and you did not ask me to touch it.

The conflict was mine, not yours. After you opened this, I landed two fixes on `main` (#192 and the narrow half of #194) and both added entries to the top of `Unreleased / Fixed` - the same few lines your podcast entry occupies. Pure CHANGELOG collision, no code overlap at all. Since I caused it, rebasing around it seemed like the wrong thing to hand back to you.

Resolution kept all three entries. No prose was edited and nothing outside CHANGELOG.md was touched. Your `for field in ("body", "text")` change is untouched.

Verified on the merged branch: 579 passed (575 on main plus your 4), `ruff check .` clean.

The review itself: this is a good fix and a well-written one. Trying `body` first so existing Podcast Index feeds cannot change behavior is the right ordering, and the fourth test - a `segments[].text` array of blank strings still falling through rather than returning an empty transcript - is the one I would have asked for if it had not been there. Naming the specific hosts that ship the Whisper spelling in the docstring is also the kind of thing that saves the next person a bisect.

One thing I deliberately did not change, so you can weigh in: the loop tries `body` across all segments first, then `text`, so a hypothetical mixed feed would yield only the `body` segments rather than an interleaving. I think that is correct - a mixed array is far more likely to be a malformed export than a real schema, and silently stitching two spellings together would hide it - but you have seen more of these files than I have.

Merging as soon as CI is approved (the push queued a new run, and fork PRs need a manual approval on this repo). Thank you for the fix.

PR #198, opened 2026-08-09
PostCompact: say whether a run propagated anything, not just that it ran

waqasgameswing-lgtm:
## The problem

`log_run "completed"` records `duration_sec` and `exit_code`, so these two runs are byte-identical in `.claude-runs/YYYY-MM-DD.jsonl`:

- one that created a dev log, updated a project note and appended to the daily note
- one that read the summary, decided nothing was new, and never touched the vault

```json
{"run_id":"...","status":"completed","ts":...,"duration_sec":43,"exit_code":0}
```

The second is not an edge case. The prompt says *"If the summary contains nothing vault-worthy, exit without making any changes"*, and that is the common outcome whenever a session already recorded its own work by hand before compacting. So `exit_code: 0` was never evidence of a write, and the run log could not answer the question people open it to ask.

I hit this checking whether a compaction had actually propagated. The log said `completed, exit_code 0`; only diffing the vault's git history across the run's start and completion timestamps showed it had written nothing at all. The observability added so a run that declined to act would not be silent stops one level short: it records that the agent *ran*, not that it *did* anything.

## The change

Adds `files_changed` to the completed line:

```json
{"status":"completed","duration_sec":43,"exit_code":0,"files_changed":0}   // propagated nothing
{"status":"completed","duration_sec":51,"exit_code":0,"files_changed":3}   // wrote three notes
```

**Counted as file mtimes against a marker created at t0, not as a `git status --porcelain` delta.** A vault under the Obsidian Git plugin auto-commits on a timer; a commit landing mid-run drops the dirty count back to zero while leaving mtimes untouched, so a status delta would report "wrote nothing" for a productive run — precisely the failure this field exists to prevent. mtimes also need no git at all, so vaults that are not repos are covered.

Dot-prefixed names are pruned at every level, which keeps `.git`, `.obsidian`'s constantly-rewritten workspace state, and the hook's own `.claude-runs` and `.claude-lock` out of its own measurement.

## The part that is not obvious

A raw count is not enough on its own, and it took a real run against a scratch vault to see why.

**The headless agent this hook spawns is a full Claude Code session, so it loads this same skill**, and its own `SessionStart`/`SessionEnd` hooks write to whatever `OBSIDIAN_VAULT_PATH` points at while the propagation is running — boards, rollups, an index, a memory mirror. Pointing the hook at an empty scratch vault:

| written by | files |
| --- | ---: |
| the propagation itself (daily note, project note, dev log) | 3 |
| the spawned session's own machinery | 216 |

Against a live vault the same effect showed up as 8 files touched in 6 otherwise idle minutes, every one of them generated. A raw count therefore never reaches 0, and the field would read "8" for a run that propagated nothing while looking authoritative.

Hence `OBSIDIAN_BG_COUNT_IGNORE`: an extended regex matched against vault-relative paths, excluded from the count. **Unset by default** — the generated set is a property of a vault's layout, and guessing it for someone else's vault would be worse than counting everything. Documented in `hooks/postcompact.hook.example.json` next to the other env vars, with the warning that daily/project folders must stay *out* of the pattern since a write there is the signal.

## How to read the field

- **`0` is the trustworthy end**: nothing was written, by anyone.
- A nonzero value is a **direction, not an attribution** — any other writer active in the same window inflates it, and it can never prove this particular agent wrote a given file.
- It is **omitted entirely** when the count is unavailable (no marker, malformed regex), so it is never a guess. `grep` exits 1 when it filtered everything out and >=2 on a bad pattern; conflating those would turn a typo into a confident, wrong zero.

## Tests

Four tests added to `tests/test_bg_agent_hook.py`, in the existing stub-`claude` style:

- no-op run logs `files_changed: 0` (and the hook's own run log does not leak into it)
- a run that writes two notes logs `2`
- `OBSIDIAN_BG_COUNT_IGNORE` excludes generated paths and keeps the real one
- a malformed pattern omits the field rather than reporting `0` — written differentially, so it also asserts a *valid* pattern still produces a count, otherwise it would pass against any build that never emits the field

**All four fail on `main`.** Full suite: 579 passed.

Two small fixes to the shared `_run_hook` helper came with it: `OBSIDIAN_BG_COUNT_IGNORE` is now cleared alongside the other vars so a developer's exported environment cannot change results, and the stub takes explicit paths to write rather than a count — a `seq 1 $n` loop looks equivalent but BSD `seq 1 0` counts *down* and yields `1 0`, so the write-nothing case silently wrote two files and read as a hook bug.

## Not included

Nothing here changes when or whether the agent runs, what it is allowed to touch, or the `async`/detach behaviour of the spawn.

### DIFF
--- a/hooks/obsidian-bg-agent.sh
+++ b/hooks/obsidian-bg-agent.sh
@@ -21,6 +21,11 @@
 # To disable again: clear OBSIDIAN_BG_AGENT_ENABLED (the gate below makes that enough).
 #
 # Optional:
+#   - OBSIDIAN_BG_COUNT_IGNORE is an extended regex of vault-relative paths to
+#     leave out of the completed line's files_changed count. Set it to whatever
+#     your vault regenerates on its own (boards, rollups, an index, a memory
+#     mirror), or files_changed never reaches 0 and cannot tell a no-op run
+#     apart from a productive one. Unset by default.
 #   - CLAUDE_VAULT_PROPAGATION=1 lets the origin project's CLAUDE.md steer
 #     propagation. If set, and the compacting project has a "## Vault
 #     propagation hints" section in its CLAUDE.md, that section (only) is
@@ -30,7 +35,9 @@
 # Logs:
 #   - $TMPDIR/obsidian-bg-agent-$(id -u).log - stdout/stderr, mode 0600
 #   - $VAULT/.claude-runs/YYYY-MM-DD.jsonl - one JSONL line per run outcome
-#     (early-exit reason, or starting + completed with duration and exit code)
+#     (early-exit reason, or starting + completed with duration, exit code and
+#     files_changed - see OBSIDIAN_BG_COUNT_IGNORE above, without which that
+#     count is dominated by files the vault regenerates on its own)
 
 VAULT="${OBSIDIAN_VAULT_PATH:-}"
 [[ -z "$VAULT" ]] && exit 0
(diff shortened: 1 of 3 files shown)


waqasgameswing-lgtm:

| pair | conflicts in |
| --- | --- |

Issue #183, opened 2026-07-31
MCP server crashes on startup: unpinned 'mcp' dependency resolves to breaking mcp 2.0.0 (fastmcp removed)


jackiepan99:
## Summary

`.claude-plugin/plugin.json` launches the MCP server with an unpinned `mcp` dependency:

```json
"mcpServers": {
  "obsidian-second-brain": {
    "command": "uv",
    "args": ["run", "--with", "mcp", "python", "${CLAUDE_PLUGIN_ROOT}/integrations/obsidian-mcp-server/server.py"]
  }
}
```

`uv run --with mcp` resolves to the latest PyPI release, which is now `mcp==2.0.0`. That release restructured the SDK and removed `mcp.server.fastmcp` (the `FastMCP` class now lives at `mcp.server.mcpserver.MCPServer` instead). `server.py` still does:

```python
from mcp.server.fastmcp import FastMCP
```

so the server subprocess crashes immediately on startup:

```
Traceback (most recent call last):
  File ".../integrations/obsidian-mcp-server/server.py", line 25, in <module>
    from mcp.server.fastmcp import FastMCP  # noqa: E402
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ModuleNotFoundError: No module named 'mcp.server.fastmcp'
```

In Claude Code this surfaces only as a generic reconnect failure (`Failed to reconnect to plugin:obsidian-second-brain:obsidian-second-brain: -32000`), which gives the user no clue that the actual cause is an upstream dependency break — took a bit of digging to trace back to this.

## Repro

```bash
uv run --with mcp python -c "from mcp.server.fastmcp import FastMCP"
# ModuleNotFoundError: No module named 'mcp.server.fastmcp'
```

## Confirmed workaround

Pinning below 2.0 restores the old API and the server starts cleanly:

```bash
uv run --with "mcp<2" python -c "from mcp.server.fastmcp import FastMCP"  # works
```

## Suggested fix

Pin the `mcp` dependency in `plugin.json`'s launch args (e.g. `"mcp<2"`, or ideally a specific known-good version), so a future `mcp` release can't silently break every install again. Same applies to `.cursor-plugin`, `.grok-plugin`, `.devin-plugin`, etc. if they use the same unpinned `--with mcp` pattern, and to `integrations/obsidian-mcp-server/live_test.py`'s documented run command.

[...]

PR #216, opened 2026-08-19
Add minimax adapter for Mavis (MiniMax Code) Plugin V1

frandev94:
Add minimax adapter for Mavis (MiniMax Code) Plugin V1 packaging.

## Summary

Add a \minimax\ adapter to \scripts/build.sh\ that emits a Mavis (MiniMax Code) Plugin V1 package under \dist/minimax/\. Mavis users can now run:

\\\ash
bash scripts/build.sh --platform minimax
\\\

and copy the resulting layout to \~/.minimax/plugins/obsidian-second-brain/\ to install.

## What this emits

\dist/minimax/\ mirrors the Plugin V1 spec at https://agent.minimax.io/docs/code/agents/plugins:

- \.minimax-plugin/plugin.json\ - V1 manifest with all required fields plus \displayName\ (optional but supported)
- \obsidian-second-brain.mcp.json\ - stdio MCP server config with \mcp<2\ pin (per #183)
- \icon.png\ - from upstream \media/icon.png\ if present, else 1x1 fallback
- \placeholder\ - write-tool anchor (per the V1 spec convention)
- \integrations/obsidian-mcp-server/{server.py,vault_ops.py,README.md}\ - verbatim copies of the upstream Python MCP server
- \skills/obsidian-second-brain/SKILL.md\ - verbatim copy of the upstream SKILL.md

The \OBSIDIAN_VAULT_PATH\ env value is templated as \<VAULT_PATH>\ so the emitted package has no user-specific paths; each Mavis user substitutes their vault at packaging time.

## Why this is platform-neutral

- The adapter does not read upstream \pyproject.toml\; the manifest \ersion\ is \

### DIFF
--- a/adapters/OWNERS.md
+++ b/adapters/OWNERS.md
@@ -1,7 +1,7 @@
 # Platform owners
 
-Seven builds are compiled from one source tree. One person maintains all seven
-and can only actually run two of them.
+Eight builds are compiled from one source tree. One person maintains all seven
+of the original and can only actually run two of them.
 
 That is not a staffing complaint, it is the root cause of a specific class of
 bug in this repo. Five separate audit findings (B3, B19, B20, S15, S23) were the
@@ -24,6 +24,7 @@ So platform ownership is open. A platform owner is the person who would notice.
 | hermes | unclaimed | [@Litash](https://github.com/Litash) |
 | opencode | unclaimed | [@Litash](https://github.com/Litash), [@MPZ-00](https://github.com/MPZ-00) |
 | pi | unclaimed | |
+| minimax | [@fcojg](https://github.com/frandev94) | |
 <!-- owners:end -->
 
 "Previously contributed" is credit for work already merged, taken from the git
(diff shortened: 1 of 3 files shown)


eugeniughelbur:
Saw this before it closed. A MiniMax/Mavis build is exactly what #171 exists for - all seven current platform builds are open for owners, and an eighth is welcome on the same terms: the owner runs the platform's own agent against the build and reports what actually resolves at runtime.

If it closed because the adapter needed more work than expected, no problem. If you use Mavis day to day and want to finish it, reopen (or open fresh against current main) and I'll review properly. One request for the reopen: run the build output against a real Mavis install and paste what resolves - that runtime half is what makes a platform build real (see the per-platform reports in #171).

frandev94:
The earlier PR was premature.
I opened it without having actually exercised the adapter against a live Mavis install, so what I shipped was a build that compiled cleanly and was probably broken in exactly the way your comment describes, closing it was the right call.
I do use Mavis day to day, so the runtime report is owed. If it holds, the next PR can carry a final commit putting me in the Owner column of OWNERS.md for minimax; if not, the row stays unclaimed. I'll come back when there's something concrete to show.

PR #223, opened 2026-08-21
eval: complete end-to-end behavior eval — synthesis/contradiction-weighted cases, judge-model separation

### LINKED ISSUE #197: Eval: measure agent behavior with vs without the vault, not just retrieval
## The gap

`scripts/eval/` measures **retrieval**: given a question, does the right note come back. That is reproducible and honest, and the numbers in `BENCHMARK.md` / `BASELINE.md` are something anyone can re-run on the synthetic corpus.

What it does not measure is **behavior**: with the vault loaded, does the agent actually give a better answer than the same agent with no vault at all.

Those are different claims. Recall@10 tells you the retrieval layer works. It does not tell you the second brain was worth building.

## Prior art worth reading

`TencentCloud/TencentDB-Agent-Memory` publishes a PersonaMem result for its memory layer: 48% without memory, 76% with (as of 2026-08, github.com/TencentCloud/TencentDB-Agent-Memory). Whatever one thinks of the number, the *shape* of the claim is the one users want answered, and it is the shape this repo currently cannot make.

## Proposed scope

Build an end-to-end eval on top of the existing synthetic 300-note corpus in `scripts/eval/corpus.py`, so nothing private is involved and anyone can reproduce it.

1. Author a set of questions that are only answerable from the corpus (facts, decisions, relationships, cross-note synthesis, and contradictions the vault should have reconciled).
2. Run each question twice: once with vault retrieval available, once with the model alone.
3. Grade both answers against a written rubric with an LLM judge, blind to which condition produced which answer.
4. Report the delta, the per-category breakdown, and every question where the vault made the answer *worse*. That last bucket is the interesting one and should not be dropped from the report.

## Why the failures matter more than the headline

A single "with memory is better" percentage is marketing. The useful output is the category breakdown: if the vault helps on entity recall and hurts on synthesis, that is a concrete direction for the ranking and preamble work, not a number for a README banner.

## Non-goals

- No real vault data. Synthetic corpus only.
- Not a replacement for the retrieval eval. It sits next to it.
- No claim shipped to the README until the harness is reproducible by a third party from a clean checkout.

## Open design question

Does a behavior benchmark for a single-user vault survive contact with reality, or does it degrade into synthetic persona testing that proves nothing? Worth arguing in this thread before anyone writes the harness. A negative result here (the honest answer being "this cannot be measured credibly at this scale") is a legitimate outcome and should be documented rather than quietly abandoned.

AaronProbha18:
## What this PR does
Completes the end-to-end behavior eval from #197: `behavior_eval.py` runs
each question twice (vault-on vs. vault-off, both via `research.lib.grok`),
grades both blind with a separate judge model (`research.lib.gpt`, new in
this PR) against the corpus's known `answer_key`, and reports the overall
delta, a per-category breakdown, and a complete, never-truncated list of
cases where the vault made the answer worse. The behavior case mix in
`corpus.py` is now weighted toward synthesis and contradiction (60 cases:
fact 20% / decision 15% / relationship 10% / synthesis 30% /
contradiction 25%) instead of the fact-dominated (~48%) set from the prior
PR, and a startup guard refuses to run if the answer and judge models
resolve to the same provider+model.

## Type of change
- [ ] 📜 New slash command
- [x] 📚 Documentation update
- [ ] 🧹 Refactor / cleanup

## Related issues
Closes #197

## How I tested it
`pytest tests/test_behavior_eval.py` - keyless, zero network access, both
API keys explicitly deleted in the `vault_env_only` fixture. Covers: case
determinism and schema, category-mix bounds against the target ratio, the
answer/judge model-separation guard (including that it fires before any
case executes), judge JSON parsing (well-formed, prose-wrapped, malformed,
missing fields, call failure, missing-key `SystemExit` propagation),
`_ab_order` reproducibility (pinned against its actual hash), and
`evaluate()` end-to-end with all LLM calls mocked (positive delta,
all-regressions, zero delta, all-unjudged, missing cases file).
Also ran `uv run python scripts/eval/corpus.py --manifest` to confirm the
new `behavior` set and mix percentages print correctly and the hash
includes them.
Have not yet run a real, keyed `behavior_eval.py` pass end-to-end against
live `XAI_API_KEY`/`OPENAI_API_KEY` - that's a manual, opt-in run per the
harness's own design (never a CI gate), not part of this PR's automated
verification.

- [ ] My changes follow `references/ai-first-rules.md`
- [x] Frontmatter includes `ai-first: true`, `type:`, `date:`, `tags:` *(unchanged from existing `corpus.py` generation, not modified by this PR)*
*(N/A - this PR touches eval tooling only, not a vault-writing command.)*

## Checklist
- [x] I updated the README if user-facing behavior changed
- [ ] I updated `CHANGELOG.md` (under "Unreleased") if applicable

## Screenshots / output (if visual or producing notes)
`uv run python scripts/eval/corpus.py --manifest` output (paste actual
output before merge), showing the `behavior` set count and the
`behavior mix   fact 20.0%, decision 15.0%, relationship 10.0%,
synthesis 30.0%, contradiction 25.0%` line alongside the sha256 hash.

## Review notes
- `BASELINE.md` / `BENCHMARK.md` are untouched, per the issue's non-goal -
  no numbers from this harness are promoted there.
- A real scored run needs both `XAI_API_KEY` and `OPENAI_API_KEY`; there is
  no key-free fallback for answering or judging (only `--generate`'s
  corpus sampling is keyless).
- `ANSWER_PROVIDER`/`JUDGE_PROVIDER` are currently hardcoded to `"grok"`/
  `"gpt"` and can't collide through normal CLI use today - the guard is
  defensive against future changes to that wiring, not live protection
  against picking the same model name via `--answer-model`/`--judge-model`
  alone (provider is what's compared, not just the model string).


eugeniughelbur:
@AaronProbha18 merged, thank you - the design points from the issue are all in and the tests hold with every LLM call mocked. I read it in full after the merge and fixed four things directly on `main` in 4a8f3d8 rather than round-tripping them, so you can see the reasoning in the diff:

1. **A retrieval miss was dropped from the delta.** `_answer_with_vault()` returned `None` on empty context and `evaluate()` filed the case as "answer generation failed". An empty retrieval is the vault failing hardest, and exactly what the regression bucket exists to show. The vault-on arm now answers with no notes and is judged like any other case; each case carries `retrieval_empty` and the summary counts `retrieval_misses`.
2. **The judge was not blind in practice.** The vault-on prompt said "using ONLY the notes below", so those answers said "the notes state" and the vault-off ones never did. Both arms now get the same rule against mentioning notes, sources or context.
3. **`BENCHMARK.md` corpus hash.** Your `build()` change moves it from `811cc66735a81d13` to `5773dc7f39b0c3b0`. I generated both corpora and ran `retrieval_eval.py --mode lexical` on all three sets: identical to three decimals, so the lexical table stands. Hybrid rows were not re-run (no Ollama here) and the doc says so. For the record, the hash in the doc was already stale before your branch: the future-agent rename moved it and nobody updated the file. You fixed a drift you did not cause.
4. **CHANGELOG entry** under Unreleased, and the startup guard's docstring now says it cannot fire from the CLI while the providers are constants.

Two things I left as they are: `gpt.call()` passes `max_tokens`, which will need `max_completion_tokens` if anyone points `GPT_JUDGE_MODEL` at an o-series model; and `_ab_order()` keyed on file index means reordering the cases file changes the labeling (documented, fine). If you run a real keyed pass, post the output here - the regression list is the part I want to read.

Issue #159, opened 2026-07-22
[Bug] Vault search silently drops all 1-2 character CJK words: the len(t) > 2 term filter in vault_ops.py discards most Chinese/Japanese queries, and the fallback repeats the same filter


SylvesterTee:
### Which command broke?

Other / not sure - the `obsidian_search` MCP tool, and anything else calling `vault_ops.search()` (`/obsidian-find`, the Search Before Write step in `references/write-rules.md`).

### What happened?

On a 1,142-note Traditional Chinese vault, search returns **zero results** for common 2-character Chinese words that are all over the vault. Words of 3+ characters work fine, as does ASCII.

| Query | Chars | Result | Ground truth (ripgrep) |
|---|---|---|---|
| `系統` ("system") | 2 | **0 results** | 6,018 occurrences across 859 files |
| `知識管理` ("knowledge management") | 4 | 2 correct results | - |
| `TSMC` | 4 (ASCII) | 5 correct results | - |

A word appearing 6,018 times is invisible to search, while a 4-character phrase in the same vault resolves correctly.

### Root cause

`integrations/obsidian-mcp-server/vault_ops.py:340`:

```python
terms = [t for t in re.split(r"\W+", query.lower()) if len(t) > 2 and t not in _STOPWORDS]
```

Two effects compound:

1. Python 3's `\w` is Unicode-aware, so **CJK ideographs are word characters** and `\W+` never splits a Chinese phrase. A CJK query arrives as a single token.
2. `len(t) > 2` is calibrated for English noise words (`a`, `is`, `of`). But **in Chinese, Japanese and Korean, two characters is the single most common full-word length** - 系統, 資料, 会議, 設備. The filter discards the majority of real CJK queries.

The fallback at L341-344 cannot rescue this, because it repeats the same length filter:

```python
if not terms:
    # Query was all stopwords/short tokens - fall back to the raw terms so a
    # search like "the who" still returns something rather than nothing.
    terms = [t for t in re.split(r"\W+", query.lower()) if len(t) > 2]   # same filter
if not terms:
    return []
```

The safety net has the same hole as the thing it is catching, so the call returns `[]`.

Longer CJK phrases do survive the filter, but they stay one un-segmented token, so they only match as an exact substring - a paraphrased question never matches.

### What did you expect to happen?

`系統` should return notes containing 系統, the same way `system` returns notes containing "system".

### Steps to reproduce

[...]

PR #228, opened 2026-08-25
Fix /notebooklm slug for non-ASCII topics

### LINKED ISSUE #227: [Bug] /notebooklm slug drops Cyrillic (and any non-ASCII script) topics
## Which command broke?
Other / not sure - /notebooklm isn't in the dropdown above, but the file is `scripts/research/notebooklm.py`.

## What happened?
Ran `/notebooklm "Идеальный Agentic PKM фреймворк"` (a Russian-only topic). `slugify()` in `scripts/research/notebooklm.py` only keeps `[a-z0-9\s-]`:

```python
def slugify(text: str) -> str:
    text = text.lower()
    text = re.sub(r"[^a-z0-9\s-]", "", text)
    text = re.sub(r"\s+", "-", text.strip())
    return text[:80]
```

This strips Cyrillic (and any other non-Latin script) entirely, so the slug collapses to empty or near-empty. It's used both for the saved note's filename and the Gemini File Search store's `display_name`:

```python
slug = slugify(topic)
...
store = client.file_search_stores.create(config={"display_name": f"vault-{slug}-{today}"})
...
note_path = NOTEBOOKLM_DIR / f"{today} - {slug}.md"
```

## What did you expect to happen?
A saved note with a non-empty, readable filename, e.g. `Research/NotebookLM/2026-08-24 - идеальный agentic pkm фреймворк.md`.

## Steps to reproduce
1. Pull latest main (`ce2b5e8`)
2. Run `uv sync`
3. `python3 -c "import sys; sys.path.insert(0, 'scripts'); from research.notebooklm import slugify; print(repr(slugify('Идеальный Agentic PKM фреймворк')))"`
4. Output is `''` (fully Russian topic) or a near-empty fragment (mixed-language topic), instead of a usable slug. A fully-Russian topic run end to end would try to write `Research/NotebookLM/YYYY-MM-DD - .md`.

## Skill version (release tag or commit SHA)
ce2b5e8

## Operating system
Other

(Not OS-specific - pure string handling, reproduces identically on macOS/Linux.)

## Relevant logs / error output
No crash - the bug is silent. The saved note just gets an empty/near-empty, potentially colliding filename instead of erroring.

## Confirmation

konsone:
## Summary

`notebooklm.py`'s private `slugify()` only kept `[a-z0-9\s-]`, so any non-Latin-script topic (Russian, etc.) produced an empty or near-empty slug, used for both the saved note's filename and the Gemini File Search store's `display_name`.

The repo already has a Unicode-aware `slugify()` in `lib/vault.py` (used by `youtube_extract.py`) that folds via `\w` (Unicode word characters), so non-Latin scripts survive intact. This swaps the private copy for that one instead of adding a fourth reimplementation of the same normalization - the class of bug this exact file has hit three times before via `vault_scan`'s tokenizer duplication (#159/#188/#192/#212).

## Changes

- Remove `notebooklm.py`'s private `slugify()` (and the now-unused `import re`).
- Import `slugify` from `.lib.vault` instead.
- Add the `or "untitled"` fallback already used by `lib/vault.py::filename_for`, so a symbol/emoji-only topic still gets a non-empty filename instead of `YYYY-MM-DD - .md`.

No behavior change for ASCII topics - `lib/vault.py::slugify` also truncates at 80 chars by default, same as the removed version.

Closes #227

## Verification

- `uvx ruff check scripts/research/notebooklm.py scripts/research/lib/vault.py` - clean
- `uv run pytest -q` - 607 passed
- Manual:
  ```
  slugify("Идеальный Agentic PKM фреймворк") -> "идеальный agentic pkm фреймворк"
  slugify("🚀🚀🚀") or "untitled" -> "untitled"
  slugify("Smart Home automation ideas") -> "smart home automation ideas"
  ```

### DIFF
--- a/scripts/research/notebooklm.py
+++ b/scripts/research/notebooklm.py
@@ -19,7 +19,6 @@ from __future__ import annotations
 
 import argparse
 import json
-import re
 import shutil
 import sys
 import tempfile
@@ -32,17 +31,11 @@ from google.genai import types
 
 from .lib.config import NOTEBOOKLM_MODEL, VAULT_PATH, get_required
 from .lib.gemini import GEMINI_DEFAULT_MODEL, MODEL_LADDER
+from .lib.vault import slugify
 from .lib.vault_terms import topic_terms
 
 MAX_BUNDLE_NOTES = 12
 NOTEBOOKLM_DIR = VAULT_PATH / "Research" / "NotebookLM"
-
-
-def slugify(text: str) -> str:
-    text = text.lower()
-    text = re.sub(r"[^a-z0-9\s-]", "", text)
-    text = re.sub(r"\s+", "-", text.strip())
-    return text[:80]
 
 
 def _vault_scan_dirs() -> list[str]:
@@ -192,7 +185,7 @@ def run(topic: str) -> int:
         return 1
 
     today = datetime.now().strftime("%Y-%m-%d")
-    slug = slugify(topic)
+    slug = slugify(topic) or "untitled"
 
     print(f"=== /notebooklm: {topic} ===", file=sys.stderr)
     print(f"Vault baseline: {len(hits)} notes", file=sys.stderr)


eugeniughelbur:
First review. The private ASCII-only `slugify` is the bug in #227. Replacing it with `lib.vault.slugify` (and `or "untitled"`) is the right fix, not a fourth copy.

One file, +2/−9. No smoke check ran on this PR. Mergeable is blocked. I am not merging it.

PR #231, opened 2026-08-27
Add /obsidian-merge: merge near-duplicate notes from /obsidian-health

### LINKED ISSUE #220: [Feature] Merge command for near-duplicate notes found by /obsidian-health
## What kind of feature?

Improvement to an existing command

## What problem would this solve?

`/obsidian-health` already detects duplicate/near-duplicate notes and surfaces them in its audit, but that's where it stops — resolving a flagged duplicate is entirely manual. For a vault that grows via repeated `/obsidian-ingest` runs, duplicates recur, and the fix-it step is the part that keeps getting skipped because it's tedious, not because it's hard to decide.

## Proposed solution

A command that takes a pair (or the pairs from the latest health report), shows a diff, and on confirmation merges content into the canonical note, redirects/aliases the old title, and removes/archives the old file. `--dry-run` to preview without touching anything.

## Alternatives you considered

Leave it manual — works today, but is exactly the step people skip, which is presumably why larger vaults report the same duplicates run after run. Could this land as a `--merge-duplicates` flag on `/obsidian-health` instead of a new command? Open to whichever fits the existing surface better.

## How does this interact with the AI-first vault rule?

Produces vault notes - will follow ai-first-rules.md

## Additional context

Touches existing notes (merge + alias) rather than creating new ones; no new note type introduced.

konsone:
## What this PR does

Implements `/obsidian-merge` per your design in #220: `scripts/merge_notes.py` does the mechanical half of merging a near-duplicate pair (frontmatter union, alias folding, replacing the retired note with a `type: redirect` stub), and `commands/obsidian-merge.md` composes the merged body (the judgment part - contradictions, provenance) and drives dry-run/confirm/apply.

## Type of change

- [x] 📜 New slash command
- [ ] 📚 Documentation update
- [ ] 🧹 Refactor / cleanup

## Related issues

Closes #220

## ⚠️ CI will show 3 failing tests - expected, please read before reviewing

Per your comment ("I will do the `SKILL.md` and README wiring on merge"), this PR intentionally does **not** touch `SKILL.md` or `README.md`. Three tests fail as a direct, unavoidable consequence: `test_doc_roster.py::test_every_command_is_in_skill_and_readme`, `test_doc_roster.py::test_headline_count_matches_filesystem`, and `test_setup_and_manual_drift.py::test_every_command_file_is_reachable_from_skill_md` - all three are cross-reference checks that `obsidian-merge` is missing from `SKILL.md`/`README.md`, which is exactly the wiring you said you'd add. Full suite: 677 passed, 3 failed (baseline on `main` is 670 passed, 0 failed; 670 + 7 new tests in `test_merge_notes.py` = 677). They should go green as soon as your wiring lands, whether that's a follow-up commit on this branch or on `main` after merge - happy to add it myself if you'd rather it ship in this PR, just say so.

## Changes

- `scripts/merge_notes.py` (new) - the mechanical merge/redirect script:
  - `compute_merge()` is the single code path both dry-run (`preview()`) and `--apply` (`apply_merge()`) call, so the preview can never show something `--apply` doesn't actually write.
  - `merge_frontmatter()`: union of keys; on a value conflict the canonical note wins and the retired note's value is recorded under a new `merged_from:` block (with the retired note's path); equal values and union-only fields never produce `merged_from` noise. The retired note's title is folded into the canonical's `aliases:`, skipped when it's identical to the canonical's own title (e.g. two same-named notes in different folders).
  - `--from-health` imports `check_duplicates`/`load_vault`/`load_vault_config` directly from `vault_health` (same pattern `heal_links.py` already uses) rather than shelling out - there's no persisted health-report file anywhere in this repo, so "the latest health report" means running the scan live. A duplicate group of more than 2 files is never auto-paired; it's reported, and merging it needs explicit `--canonical`/`--retire`, since nothing in the health check says which note should survive.
  - The merged **body** is deliberately not composed here - that's judgment (what actually contradicts between two notes), the same split `heal_links.py` (mechanical) draws against `triage_links.py` (AI judgment). The caller (the command) composes it and hands it in via `--merged-body-file`, required on every run including dry-run, so the preview is always the literal text `--apply` would write.
  - `dump_frontmatter()`: there's no YAML serializer anywhere in this repo (`yaml.dump`'s default style doesn't match this project's hand-tuned frontmatter conventions), so this writes clean text directly - bare where safe, quoted where YAML would otherwise misparse it (a bare `[[wikilink]]` is invalid flow-sequence syntax, so `redirects_to` and any other wikilink value are always quoted).
- `commands/obsidian-merge.md` (new) - takes two paths or `--from-health`; if `$ARGUMENTS`' ordering isn't obviously the intended canonical/retired pair, or a `--from-health` group has 3+ files, it asks rather than guesses. Composes the merged body per `references/ai-first-rules.md` (preamble naming both source notes, both provenance trails kept, contradictions listed not resolved), runs the dry run, requires explicit confirmation before `--apply`.
- `references/ai-first-rules.md` - a "Redirect exception" bullet under `## Documented exceptions` plus the `type: redirect` schema (placed there rather than under `## Type Schemas`, matching how `adr`/`brainstorm`/`podcast` are physically placed in that same section - redirect is the one true relaxation among them, since it deliberately carries no preamble, no wikilinks-everywhere, no sources).
- `tests/test_merge_notes.py` (new, 7 tests) - your explicit ask (dry run against two synthetic notes writes nothing) plus: apply writes the redirect stub + merged frontmatter + alias, conflict resolution (canonical wins, loser recorded with its source path), no-conflict means no `merged_from` noise, `--from-health` resolves a real pair end-to-end and refuses an ambiguous 3-file group, and a missing `--merged-body-file` errors before any write.
- `CHANGELOG.md` - entry under Unreleased.
- `docs/index.html`, `docs/commands/obsidian-merge.html` - regenerated via `scripts/build_site.py`. This is mechanical (`commands/` is its only declared source), so I included it rather than leaving the site out of sync with `commands/`; not part of the `SKILL.md`/README territory you reserved.

**Not touched**: `SKILL.md`, `README.md` (your territory per your comment), `hooks/validate-ai-first.sh`, `references/write-rules.md`'s existing "Archiving" section (the `_archived_`-filename convention there is a separate, pre-existing mechanism for a different purpose - didn't touch or try to reconcile it with the new `type: redirect` approach this issue specifies).

## How I tested it

- `uv run pytest tests/test_merge_notes.py -q` - 7 passed.
- `uv run pytest -q` (full suite) - 677 passed, 3 failed (see the CI note above).
- `uvx ruff check .` - clean.
- Manual: ran `merge_notes.py` by hand against a synthetic vault for both the dry-run and `--apply` paths, confirmed the redirect stub's wikilink resolves to the canonical note's actual title and that re-running with `--apply` after a dry-run produces byte-identical output to what the dry-run previewed.

- [x] If I added or changed a command that writes to the vault, the produced note has a `## For future Claude` preamble *(the canonical/merged note does; the retired note becomes `type: redirect`, this PR's one documented exception - see `references/ai-first-rules.md` § Documented exceptions, "no preamble" is deliberate for a note whose only job is to point elsewhere)*
- [x] Wikilinks (`[[...]]`) used for every person, project, idea, decision referenced
- [x] External claims have recency markers (`(as of YYYY-MM, source.com)`) *(n/a here - the command composes a merge of existing vault notes, not new external research)*
- [x] Sources preserved verbatim (URLs inline, not paraphrased)

## Checklist

- [x] I updated the README or docs if user-facing behavior changed *(`docs/` regenerated; `SKILL.md`/`README.md` deliberately left to you, see above)*
- [ ] If I added a new command, I added an entry in `SKILL.md` and the README's command table *(intentionally not done - see the CI note above)*


eugeniughelbur:
First review. Implements `/obsidian-merge` to the design in #220: dry-run default, redirect stub (not delete), frontmatter union with `merged_from`, body composed by the command, `--from-health` refuses 3+ groups. Tests cover the ask. Docs site regenerated; SKILL.md and README left for maintainer wiring as requested — that is why smoke is red (3 doc-roster failures, expected). Mergeable blocked. +755/−1, 7 files.

Needs your SKILL.md + README wiring (here or right after merge) before green. I am not merging it.

eugeniughelbur:
Second pass. Still not merge-ready until `SKILL.md` and README list `/obsidian-merge` (the 3 red doc-roster tests). Smoke still red (job 98499124454).

`merge_frontmatter` treats `tags` and `aliases` as one value. If they differ, the canonical list wins and the retired list goes to `merged_from` — it does not union items. A note can lose tags unless the command author copies them by hand.

`redirects_to` is `[[` + file stem + `]]`, not the title in frontmatter. That misses when the filename and the note title differ.

I am not merging it.

eugeniughelbur:
@konsone merged as a merge commit (58e79f4), with the wiring and two fixes on top in b8c52f8. CI is green on `main`.

**Done for you:** `SKILL.md` section, README table row, and the command counts in README, CLAUDE.md and architecture.md (46 -> 47). Thank you for regenerating the docs site in the PR; `build_site --check` passed untouched.

**Two changes to `merge_notes.py`, with tests:**

1. **List fields union instead of conflicting.** `tags`, `aliases`, `related-*` present on both notes now merge as canonical's items plus the retired note's new ones (case-insensitive dedup), and do not appear under `merged_from`. Canonical-wins on a list silently dropped every tag only the retired note had, a loss a merge must not cause.
2. **Same-stem redirect is path-qualified.** For `Ideas/Project Alpha.md` merged over `Archive/Project Alpha.md`, the stub now says `redirects_to: "[[Ideas/Project Alpha]]"`. A bare `[[Project Alpha]]` is ambiguous in Obsidian and can resolve to the redirect stub itself, which is a loop. I updated the assertion in `test_from_health_resolves_a_real_duplicate_pair` to match. Different stems keep the bare link, since that is what every existing wikilink to the canonical note already uses.

One review note I did not act on: the "redirects_to uses the file stem, not the frontmatter title" finding. The stem is correct. Obsidian resolves `[[X]]` against filenames, not `title:` fields, so linking by stem is what keeps old links working. Same reasoning for the alias fold.

`/obsidian-merge` is live. #220 closed by the merge.

PR #234, opened 2026-08-30
Raise the transcript cap in /youtube and /podcast from 24k to a configurable 480k

### LINKED ISSUE #233: [Bug] /youtube and /podcast silently truncate transcripts at 24k chars - 90%+ of long content is discarded from summaries
**Which command broke:** Other / not sure (affected: `/youtube`, `/podcast`)

**What happened:**
`youtube_extract.py` line 167 and `podcast_extract.py` line 151 hardcode `TX_LIMIT = 24000  # ~6k tokens - plenty for grok-4 context`. Both "plenty for grok-4" and "~6k tokens" are stale: since the Gemini summarization path (`lib/gemini.py`) landed, the default ladder resolves to Gemini flash models with a 1M-token context, and the 24k cap now silently discards 90%+ of any interview-length content. Measured on a 3-hour Lex Fridman episode (#501): 289,911-char transcript, only the first 24k chars reached the model - the summary lost the Omarchy Quattro details, the C++/Qt/Rust transition segment, the 16-parallel-agents setup, and produced follow-ups the full transcript contradicts.

**Expected:** Summary built from the full transcript (or from an explicitly configurable limit).

**Reproduce:** Run `/youtube https://www.youtube.com/watch?v=NYFGCESmikA` (289K chars) and compare against the same run with the cap removed.

**Environment:** Linux, `GEMINI_SUMMARY_MODEL=gemini-flash-lite-latest` (forced - `gemini-2.5-flash` 404s as "no longer available to new users" on this key; also reproducible on the `gemini-2.5-flash` ladder default for keys where it still resolves).

**Version:** v0.14.0 (latest tag)

**Evidence (our runs, all $0.0000):**

| Config | Input tokens | Truncation note |
|---|---|---|
| current code (24k cap) | ~6K | yes |
| patched (480k cap) | 74,553 | none |

**Proposed fix:** Read the cap from config with a 480k default - `config.get_optional("YOUTUBE_TX_LIMIT"/"PODCAST_TX_LIMIT", "480000")` - keeping the truncation note behavior for anything still larger. PR to follow.

Confirmations: searched existing issues (no duplicate found), reproduced on latest main.

konsone:
## Summary
- The 24k-char `TX_LIMIT` in both `youtube_extract.py` and `podcast_extract.py` predates the Gemini summarization path and its 1M-token context; comments still claimed it was "plenty for grok-4 context."
- Measured on a 3-hour episode: only ~6K input tokens of a 289,911-char transcript reached the model, discarding 90%+ of the content.
- Both caps now read `YOUTUBE_TX_LIMIT` / `PODCAST_TX_LIMIT` via `config.get_optional()`, default 480k chars (~120k tokens), keeping the existing truncation-note behavior for anything still larger.

Closes #233.

## Test plan
- [x] `ruff check` on both changed files (passed)
- [x] `uv run python -c "from scripts.research import youtube_extract, podcast_extract"` (imports clean)
- [x] Measured before/after on a real 3h episode (see #233): 24k cap -> ~6K input tokens, truncated; 480k cap -> 74,553 input tokens, no truncation
- [x] CHANGELOG "Unreleased -> Fixed" entry added
- [ ] No README/SKILL.md update needed (env var only, no user-facing flag)

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>


eugeniughelbur:
First review (Compound Engineering). The default-path fix is correct and matches #233. Smoke tests passed. Nothing here should block merge.

Non-blocking: a bad `YOUTUBE_TX_LIMIT` / `PODCAST_TX_LIMIT` (`480k`, `1e6`) will traceback instead of a `SystemExit` that names the env var. The new knobs are not in `.env.example`. `/podcast` still goes to Grok, so the 480k default is a ~20x prompt vs the old 24k cap.

I am not merging this.

PR #235, opened 2026-08-30
Add free Groq-hosted Whisper fallback to /podcast transcript chain (+ Gemini-first summarization)

konsone:
Refs #233 (same flow as #234). Two commits, plus one small follow-up fix:

1. Groq Whisper fallback: `rss-transcript-tag -> groq-whisper-api (free, GROQ_API_KEY) -> whisper-api -> show-notes`. Downsample-first 32kbps mono re-encode fits ~4.5h inside the 25MB cap in one call; longer episodes split with exact-bitrate math, 10s overlap, sentence-boundary seam dedup; metadata stripped (one test episode carried an 18MB XMP blob). Episode-level all-or-nothing - providers never mixed.
2. Gemini-first summarization mirroring /youtube (GEMINI_API_KEY -> Gemini, any failure -> Grok, no key -> unchanged); Apple `?i=` trackId now resolves via a second iTunes lookup (`entity=podcastEpisode`) to an episode title instead of silently falling back to the most recent episode.
3. One test-data string in `tests/test_groq.py` used an em-dash, which `test_smoke.py::test_substitution_check_passes_on_repo` correctly flags per CONTRIBUTING's own banned-substitution-character rule. Swapped for a plain hyphen - `_norm_sentence` strips all non-alphanumerics either way, so the test still demonstrates the same thing.

How I tested it:
- `ruff check` clean on all touched files.
- 39 new offline tests (`test_groq.py` 27 + `test_podcast_resolution.py` 12): real local HTTP server drives Groq 429/200/empty/unconfigured paths over an actual socket (return value, request count, Authorization header); real ffmpeg/ffprobe drive chunk-split math; fallback-chain tests assert exact call counts; iTunes resolution tests via a fake server; Gemini/Grok branch selection tests.
- Full suite: **713 passed, 0 failed** on this branch (`uv run pytest tests/ -q`), including the substitution-character CI gate after the fix above.
- Live end-to-end (run before this PR, on the working implementation): 36-min episode (no transcript tag) -> groq-whisper-api, `transcript: 38134 chars`, cost $0.00. 147.7-min episode -> auto-split into chunks, transcript 127,360 chars, cost $0.00. Apple `?i=` URL and Spotify URL both resolve correctly to the right episode.

Do NOT squash-merge; maintainer merges. Nothing else in the repo was touched.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>


eugeniughelbur:
@konsone merged as a true merge commit (40d0b8e), not a squash, as asked. Read the whole thing before merging; the design holds and the tests are the ones I want more of (real sockets, real ffmpeg). Four follow-ups landed on `main` in b8c52f8 instead of bouncing them back:

1. **Free-tier honesty.** The 25MB fit is real, but the free tier also caps audio at 7,200 seconds per hour, so a single episode over about 2h gets a 429 and falls through, and a second long episode in the same hour does too. `commands/podcast.md` and `.env.example` now say so next to the "up to ~4.5h" claim, so nobody reads that as "free for 4.5h".
2. **`GROQ_API_KEY`, `YOUTUBE_TX_LIMIT`, `PODCAST_TX_LIMIT` in `.env.example`**, and a `config.get_optional_int()` so a cap written as `480k` exits naming the variable instead of tracebacking inside the command.
3. **CHANGELOG** for the Gemini-first summarization and the Apple `?i=` fix, which the PR's entry did not cover.
4. **Provenance string.** The note preamble said "Summarized via Grok" even when Gemini wrote it; it now names whichever model ran.

Two things I noticed and left alone, for your judgment: `_join_chunk_texts` drops one duplicated sentence per seam, but a 10s overlap in ordinary speech holds two or three complete sentences, so residual duplicates remain at seams on episodes over 4.5h (rare path, worth a note in the docstring at least); and a sentence longer than 10s that spans a cut is partial in both chunks and gets trimmed from both, so it is lost. Neither blocks anything at these episode lengths.

PR #209, opened 2026-08-15
Stop promising auto-read of _CLAUDE.md before the wiring exists

motkoning:
## What this PR does

Rewords bootstrap's closing message, which promised "Claude will read it automatically on every session" - a claim that needs three pieces of wiring (the `load_vault_context` SessionStart hook, `OBSIDIAN_VAULT_PATH` in the process environment, and a session started inside the vault), none of which a bare-clone bootstrap has. The message now leads with what the user must ensure (the agent reads `_CLAUDE.md` before working in the vault) and names the wiring that automates it. Same claim-vs-wiring gap as the write-time hook docs fixed for #171.

## Type of change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)
- [x] 📚 Documentation update

## Related issues

Closes #206

## How I tested it

- Message-only change: ran `bootstrap_vault.py` on a temp path and read the output; `tests/test_showroom.py` and `tests/test_front_door.py` pass (bootstrap behavior unchanged).
- Full suite: `uv run pytest -q` - 591 passed on this branch. `ruff check` clean.
- Grepped the tree: this was the only unconditional copy of the claim outside `dist/`.

## AI-first rule compliance (required for any vault-writing command)

- [x] My changes follow `references/ai-first-rules.md`

Console message only; no vault writes touched, so the note-shape items are n/a.

## Checklist

- [x] I searched existing issues and PRs before opening this one
- [x] My commit messages are descriptive (focus on WHY, not just WHAT)
- [x] I updated the README or docs if user-facing behavior changed (the message is the behavior)
- [x] I updated `CHANGELOG.md` (under "Unreleased")

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #243, opened 2026-09-03
Windows support: fix the write-time validator, eight test failures, and the config home split

### LINKED ISSUE #240: [Bug] validate-ai-first.sh is a silent no-op on Windows: the backslash file path never matches the vault prefix
### Which command broke?

Other / not sure (`hooks/validate-ai-first.sh`, the PostToolUse write-time validator, on a plugin install in Claude Code for Windows)

### What happened?

On Windows the validator never warns. Two deliberately bad writes (a note with no frontmatter, once via Write and once via Edit) produced no `AI-first warning` in the session, while the same script run by hand with a forward-slash path warned as expected.

Cause: Claude Code on Windows passes `tool_input.file_path` with backslashes (`C:\Users\me\Vault\Inbox\note.md`) while `OBSIDIAN_VAULT_PATH` is written with forward slashes (`C:/Users/me/Vault`). The vault-scope check is a plain string prefix match:

```bash
case "$FILE" in
  "$VAULT"/*) ;;
  *) exit 0 ;;
esac
```

so it never matches on Windows and the hook exits 0 before reading the note. The AI-first rule that `CLAUDE.md` calls enforced, including check 6 (secret material in a vault note), enforces nothing on any Windows install.

The warning JSON (`systemMessage`, `decision: block`, `additionalContext`) for a note without frontmatter, exactly as on macOS and Linux.

### Steps to reproduce

From the repo root on Windows (Git Bash), with a vault at `C:/tmp/vault` containing `Inbox/bad.md` = `# no frontmatter`:

```bash
printf '{"tool_name":"Write","tool_input":{"file_path":"C:\\tmp\\vault\\Inbox\\bad.md"}}' \
  | OBSIDIAN_VAULT_PATH="C:/tmp/vault" bash hooks/validate-ai-first.sh
# prints nothing, exit 0

printf '{"tool_name":"Write","tool_input":{"file_path":"C:/tmp/vault/Inbox/bad.md"}}' \
  | OBSIDIAN_VAULT_PATH="C:/tmp/vault" bash hooks/validate-ai-first.sh
# prints the AI-first warning JSON
```

### Skill version (release tag or commit SHA)

`main` at c722701 (plugin 0.14.0)

### Operating system

Windows 11, Git for Windows bash 5.2, Claude Code plugin install (`claude plugin install obsidian-second-brain@obsidian-second-brain`)

None: the failure mode is silence. Verified with the hook's own stderr mirror absent on the backslash run and present on the forward-slash run.

### Confirmation

### LINKED ISSUE #241: [Bug] link_graph.py crashes on Windows for titles outside cp1252; the external engine path loses its backslashes; six tests carry POSIX assumptions
### Which command broke?

Other / not sure (`scripts/link_graph.py`; also `scripts/eval/retrieval_eval.py --mode external`, and the test suite on Windows)

### What happened?

`scripts/link_graph.py` crashes on Windows for any vault whose titles carry a character outside cp1252. It prints its graph with `json.dumps(..., ensure_ascii=False)`, and on Windows a pipe's stdout defaults to cp1252, so a decomposed title such as `Gru\u0308ndung` (u + U+0308, the macOS filename form the repo's own NFC/NFD test exists for) raises:

```
UnicodeEncodeError: 'charmap' codec can't encode character '\u0308' in position 45: character maps to <undefined>
```

`PYTHONUTF8=1` makes it pass, which confirms the cause. `bootstrap_vault.py` already forces UTF-8 stdout for the same reason; `link_graph.py` and `retrieval_eval.py --json` do not.

Two smaller things surfaced by the same suite run on Windows:

- `retrieval_eval.py --mode external` splits `RETRIEVAL_EVAL_EXTERNAL_CMD` with POSIX-mode `shlex.split`, which treats a backslash as an escape, so a Windows engine path (`bash C:\Users\me\engine.sh`) loses its separators and the engine is never found (recall silently 0.0).
- Six tests carry POSIX assumptions: two assert `chmod 600` survives a rewrite (NTFS has no mode bits, it reads back 0666), three redirect `HOME` to a fake home (Python on Windows resolves `Path.home()` from `USERPROFILE` and ignores `HOME`, so the scripts read the real config or raise "Could not determine home directory"), and the background-agent test compares `$PWD` with a Windows path while Git Bash reports the temp folder as `/tmp/...`.

Full suite on Windows before any change: 583 passed, 8 failed.

`link_graph.py` to emit its JSON on any platform; the external engine path to survive on Windows; the suite to be green on Windows or to skip only what cannot apply there (the repo already does this once, in `test_note_safety.py`).

### Steps to reproduce

```bash
mkdir -p /tmp/lg/wiki
printf -- '---\ntype: note\n---\nContent.\n' > "/tmp/lg/wiki/$(printf 'Gru\xcc\x88ndung').md"
printf -- '---\ntype: project\n---\nSee [[Gründung]] for background.\n' > /tmp/lg/wiki/Hub.md
uv run python scripts/link_graph.py --path /tmp/lg > /dev/null   # Windows: UnicodeEncodeError, exit 1
PYTHONUTF8=1 uv run python scripts/link_graph.py --path /tmp/lg > /dev/null   # exit 0
uv run pytest -q   # Windows: 8 failed, 583 passed
```

### Skill version (release tag or commit SHA)

`main` at c722701 (plugin 0.14.0)

### Operating system

Windows 11, Python 3.12.10, Git for Windows bash 5.2

```
FAILED tests/test_bg_agent_hook.py::test_full_chain_spawns_agent_with_summary
FAILED tests/test_eval_ruler.py::test_generate_refuses_to_overwrite_baseline
FAILED tests/test_mcp_vault_ops_safety.py::test_write_preserves_mode
FAILED tests/test_setup_and_manual_drift.py::test_setup_preserves_the_env_file_mode
FAILED tests/test_smoke.py::test_link_graph_resolves_unicode_composition
FAILED tests/test_smoke.py::test_research_key_in_config_env_selects_paid_mode[scripts.research.research-run_paid-run_free]
FAILED tests/test_smoke.py::test_research_key_in_config_env_selects_paid_mode[scripts.research.research_deep-run_paid_deep-run_free_deep]
FAILED tests/test_smoke.py::test_retrieval_eval_external_mode
```

### Confirmation

### LINKED ISSUE #242: [Bug] On Windows the bash and Python halves resolve the config home to different folders when HOME and USERPROFILE disagree
### Which command broke?

Other / not sure (`install.sh`, `scripts/setup.sh`, `update.sh`, `scripts/quick-install.sh`, `scripts/run-command.sh`, the Telegram setup, `hooks/validate-ai-first.sh` on one side; the research toolkit, `retrieval_eval.py`, and the MCP server on the other)

### What happened?

On Windows the two halves of the toolkit can resolve `~/.config/obsidian-second-brain/.env` (and `~/.claude`) to two different folders, so a user who follows the README configures one half only, silently.

- Python's `Path.home()` reads `USERPROFILE` and ignores `HOME` (Python 3.8+). The research toolkit (`scripts/research/lib/config.py`, `source_config.py`), `retrieval_eval.py`, and the MCP server (`vault_ops.py`) all use it.
- The bash scripts read `$HOME`. Git Bash normally sets `HOME` to `USERPROFILE`, but a machine whose Windows environment defines `HOME` itself (a corporate roaming home on another drive, `H:\` here) keeps that value.

On such a machine `install.sh` and `scripts/setup.sh` write the `.env` to `H:\.config\obsidian-second-brain\`, the validator hook reads its fallback from there, while the Python tools read `C:\Users\me\.config\obsidian-second-brain\`. The classic installer also writes its hook and commands under `H:\.claude`, a directory Claude Code (which keys on `USERPROFILE` too) never reads. Nothing reports the split: research commands simply stay in free mode, the MCP server cannot find the vault from the file, the hook enforces nothing.

A second, smaller gap inside Python itself: `vault_ops.py` honors the `OBSIDIAN_ENV_FILE` override (#160), but the research loaders and the eval only read the default location, so there is no single variable that steers every half.

One rule for where the config lives on Windows, shared by bash and Python, and one override (`OBSIDIAN_ENV_FILE`) that every reader honors. Proposed: on Windows shells (Git Bash, MSYS2, Cygwin) the bash scripts use `USERPROFILE`, which is what Python and Claude Code already resolve `~` to there; macOS and Linux keep `HOME` and change nothing.

### Steps to reproduce

On Windows in Git Bash, with `HOME` set to a different drive than `USERPROFILE` (or simply `export HOME=/h/`):

```bash
uv run python -c "from pathlib import Path; print(Path.home())"   # C:\Users\me
bash -c 'echo $HOME/.config/obsidian-second-brain/.env'           # /h//.config/obsidian-second-brain/.env
```

Then `bash install.sh` writes the `.env` under `H:` and `uv run -m scripts.research.research "topic"` reports free mode with the keys in place.

### Skill version (release tag or commit SHA)

`main` at c722701 (plugin 0.14.0)

### Operating system

Windows 11 with a corporate `HOME=H:\` in the Windows environment; Git for Windows bash 5.2; Python 3.12.10

```
OBSIDIAN_VAULT_PATH not configured.
Add it to C:\Users\me\.config\obsidian-second-brain\.env
```
printed by `scripts/research/lib/config.py` while the file written by `install.sh` sits at `H:\.config\obsidian-second-brain\.env`.

### Confirmation

motkoning:
## What this PR does

Makes the plugin work on Windows the way it already works on macOS and Linux, in five commits: the write-time validator was a silent no-op there (#240), `scripts/link_graph.py` crashed on any title outside cp1252 and seven more tests failed on POSIX assumptions (#241), the bash and Python halves resolved the config and `.claude` home to different folders when `HOME` and `USERPROFILE` disagree (#242), a fourth commit hardens all three after eight rounds of adversarial review, and a fifth states the changelog's suite result in a form a flaky upstream test cannot move. One PR rather than three siblings on purpose: the changelog entries land on the same lines, and #209 showed what sibling merges do to those.

Per commit:

1. **`validate-ai-first.sh` normalizes both paths before the vault-scope check, on Windows shells only.** `/c/...` and `/cygdrive/c/...` are mapped to the drive form by hand (each runtime misreads the other's spelling), `cygpath -m` then gives the mixed form, a trailing separator is stripped, and the comparison, the `.md` gate and the excluded-folder patterns are case-insensitive there because the filesystem is; the real path is kept for I/O. On macOS and Linux the paths are compared exactly as given, so a legal backslash in a filename stays intact. The `.env` fallback drops a trailing carriage return (a file written on Windows and read by macOS or Linux bash never matched) and accepts a native backslash path in `OBSIDIAN_ENV_FILE`; a note saved with CRLF line endings is validated from a carriage-return-free copy while the Python scans (banned characters, secret material) read the real file. Bash 3.2 features only, so the macOS system bash keeps running it unchanged.
2. **`link_graph.py` and `retrieval_eval.py --json` force UTF-8 stdout** the way `bootstrap_vault.py` already does. `--mode external` splits `RETRIEVAL_EVAL_EXTERNAL_CMD` in non-POSIX mode on Windows with one layer of surrounding quotes removed, so Windows paths, quoted paths with spaces and quoted literals like `'\d+'` survive; on every platform a JSON array is accepted as the exact form for anything shell quoting cannot express. The six POSIX-assumption tests are adjusted test-side: two file-mode checks skip only the mode-bit assertion on Windows, which keeps no POSIX owner/group distinction so `chmod 600` cannot be verified through `st_mode` there (the pattern `test_note_safety.py` already uses; their other assertions still run), three home-redirection tests also set `USERPROFILE`, the background-agent test compares the path in the form Git Bash reports, and the link-graph tests decode UTF-8 explicitly. Nothing changes in effect on macOS or Linux.
3. **On Windows shells the bash scripts resolve the home as `USERPROFILE`** (through `cygpath`, with the separators flipped as the fallback), matching Python and Claude Code; elsewhere `HOME` stays the home. The research loaders and the eval honor `OBSIDIAN_ENV_FILE`, the override `vault_ops.py` and the hook already accepted, and `install.sh` / `scripts/setup.sh` write the file there when it is set, so one variable steers every half; the installers print the resolved `.claude` paths instead of `~/...`. README documents the `USERPROFILE`-based commands for a corporate-`HOME` machine (bash's `~` still follows `HOME`), the native-path requirement for `OBSIDIAN_ENV_FILE` on Windows, and the caveat that a Cygwin- or MSYS-built Python resolves its home from `HOME`. This commit is the one design call in the PR; if you would rather resolve it differently, the first two stand on their own.
4. **Review hardening** (eight rounds; the commit message lists every finding): the CRLF handling, the drive spellings, the case folding, the temp copy and the Python scans, the JSON form, the installers' override and messages, and the wording above all come from that pass, each with a test.

## Type of change

- [ ] 📜 New slash command
- [x] 📚 Documentation update
- [ ] 🧹 Refactor / cleanup

## Related issues

Closes #240
Closes #241
Closes #242

## How I tested it

- Full suite on Windows 11 (Python 3.12, Git for Windows bash 5.2), branch rebased on current `main`: **731 passed, 3 skipped, 6 failed**. The six failures are pre-existing on pristine `main` (they fail identically there, from code newer than this branch's base) and are outside this PR: `test_freshness_lint.py::test_obsidian_comments_are_invisible_not_content` (the linter reports `Boards\Work.md`, the test expects forward slashes), four `test_merge_notes.py` cases where merged frontmatter comes back without `status`/`tags`/`related-projects`, and `test_research_cjk_and_models.py::test_vault_scan_finds_cjk_topic`. One more test, `test_research_sources.py::test_cache_roundtrip`, is a timing flake (a zero-hour TTL racing the file mtime): it failed in one earlier full run of this branch and in 2 of 6 runs on pristine `main`, and passed in the final run. Happy to take the six in a follow-up.
- Before this branch, the suite on Windows at the old base had 8 failures (583 passed); those eight reproduce identically on the parent commit, which is what the first three commits fix.
- Verified to fail when reverted: with this branch's test files run against unfixed `main`, the tests that guard the code changes fail (`test_validate_hook_matches_windows_backslash_paths`, `test_validate_hook_env_fallback_uses_the_platform_home`, `test_research_config_honors_env_file_override`, `test_link_graph_resolves_unicode_composition`, `test_retrieval_eval_external_mode`) and the test-side adjustments pass, as they should.
- Live: on a plugin install in Claude Code for Windows, a deliberately bad write (no frontmatter) came back with the `AI-first warning` after this hook, and produced nothing before it; a CRLF `.env` under `USERPROFILE` resolves the vault with `OBSIDIAN_VAULT_PATH` unset.
- The Windows-only tests skip on other platforms; the POSIX forms are covered by the existing tests, which run unchanged. I could not run macOS bash 3.2 myself; the hook uses only `${var//\\//}`, `[[ =~ ]]`, `BASH_REMATCH`, `case`, `shopt nocasematch` (3.1+), `tr`, `mktemp`, `trap` and command substitution, all present in 3.2.
- Reviewed adversarially before submission with OpenAI Codex (gpt-5.6-sol, max reasoning): eight rounds. Rounds one to seven produced findings (CRLF handling in the .env fallback and in notes, the MSYS and Cygwin drive spellings, case-insensitive matching, the temp copy and the Python scans, a JSON form for the engine command, the installers honoring the override and printing resolved paths, wording), each folded into the hardening commit with a test; round eight reported no findings and "ready to submit".

- [ ] If I added or changed a command that writes to the vault, the produced note has a `## For future Claude` preamble (n/a: no command changed)
- [ ] Frontmatter includes `ai-first: true`, `type:`, `date:`, `tags:` (n/a)
- [ ] Wikilinks (`[[...]]`) used for every person, project, idea, decision referenced (n/a)
- [ ] External claims have recency markers (`(as of YYYY-MM, source.com)`) (n/a)
- [ ] Sources preserved verbatim (URLs inline, not paraphrased) (n/a)

## Checklist


eugeniughelbur (commented):
## First review (comment only — no merge)

**What is good**
- Matches the three reported bugs on current `main`: silent Windows no-op in `validate-ai-first.sh` (backslash vs slash prefix), `link_graph` / eval UTF-8 and Windows `shlex`, and bash `$HOME` vs Python `USERPROFILE` config split.
- Path normalize is gated to MINGW/MSYS/CYGWIN, so macOS/Linux keep literal backslashes in filenames.
- Tests + CHANGELOG + README for the Windows config path. Smoke tests green on this head.

**What is risky / needs a yes**
- Design call: on Windows shells, bash home becomes `USERPROFILE` (not `$HOME`). That is the right match for Python and Claude Code, but it is a behavior change for anyone who relied on a corporate `HOME`. README covers it; still want an explicit maintainer yes before merge.
- Stack with #248: this PR alone still leaves six Windows failures that #248 fixes. Prefer merge order **#243 then #248** (or one combined land), not #248 alone.

**Verdict for Eugeniu:** first-pass looks solid. Waiting on your yes to merge (and on the USERPROFILE call). Not approving or merging from this bot.

PR #257, opened 2026-09-06
Fix FRESH-3 false positives on RDF CURIEs and URL path segments

wakotaki:
## What

`scripts/freshness_lint.py` FRESH-3 (typed pointer with no mapping) rang on two kinds of text that are not pointers:

- RDF / ontology CURIEs: `owl:Class`, `rdfs:label`, `skos:broader`, `dcterms:creator`. These name terms in a vocabulary, not records in a home system, so there is nothing to map or refresh. The common prefixes now join `POINTER_IGNORE`.
- `prefix:token` segments inside URLs, e.g. Medium image links `https://miro.medium.com/v2/resize:fit:1400/format:webp/...`. The existing URL guard only inspected the regex match itself, so the segment slipped through. URLs are now dropped from the line before the pointer scan.

## Why

Running the lint on a 6,400-note research vault (maritime engineering + AI/ML, many ontology notes and clipped Medium articles) produced 139 findings, of which 126 were FRESH-3 and all 126 were these two shapes. The 13 FRESH-1 findings were genuine and useful.

## Test

`tests/test_freshness_lint.py::test_fresh3_ignores_curies_and_url_path_segments`: CURIEs and a Medium URL on the same note as an unmapped `linear:ABC-123`; only the latter rings. `pytest tests/test_freshness_lint.py` and `ruff check` pass. CHANGELOG entry under Unreleased.

Context: from a research domain fork (ECOSYSTEM.md pattern) that runs the lint nightly from a scheduler.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 
 ## [Unreleased]
 
+### Fixed
+
+- **`freshness_lint.py` FRESH-3 no longer fires on RDF CURIEs or on `prefix:token` segments inside URLs.** `owl:Class`, `rdfs:label`, `skos:broader` name terms in a vocabulary, not records in a home system, so `owl`, `rdf`, `rdfs`, `xsd`, `skos`, `foaf`, `dc`, `dcterms`, `schema`, `prov`, `sh`, `dbo` and `wdt` join `POINTER_IGNORE`; and URLs are dropped from the line before the pointer scan, because a path segment such as Medium's `/resize:fit:1400/` matched the pointer shape while the URL guard only inspected the match itself. On a 6,400-note research vault this removed 126 findings, all false; a real unmapped id (`linear:ABC-123`) still rings. Covered by `tests/test_freshness_lint.py`.
+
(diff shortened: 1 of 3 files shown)


eugeniughelbur (commented):
First review (comment only — not an approve, not a merge).

**What looks good**
- Real vault pain: 126 FRESH-3 false positives on RDF CURIEs (`owl:Class`, `rdfs:label`, …) and Medium-style `/resize:fit:1400/` URL path segments; the 13 FRESH-1 hits stayed useful.
- Two precise fixes: common ontology prefixes join `POINTER_IGNORE`, and URLs are stripped from the line before the pointer scan (the old URL guard only saw the match itself).
- Test pins both exemptions next to a real unmapped `linear:ABC-123` that still rings.
- CHANGELOG under Unreleased is clear. Diff is small (3 files, +31/−2).

**Notes (non-blocking)**
- Smoke has not run yet on this fork head (`mergeable_state: blocked`, no check runs). Likely first-time contributor workflow approval — please kick CI when you can.
- `schema` (and short prefixes like `dc`) in the ignore set will also silence a home-system pointer if anyone ever used those names that way. Ontology use is the common case; call out only if you want a vault-local allowlist later.

**Verdict:** clean lint fix. Waiting on owner yes to merge once Smoke is green.

PR #264, opened 2026-09-08
Semantic index: make the per-note embedding cap configurable (OBSIDIAN_EMBED_MAX_CHUNKS)

### LINKED ISSUE #263: [Feature] Make the per-note embedding cap configurable (OBSIDIAN_EMBED_MAX_CHUNKS)
### What kind of feature?
Improvement to an existing command — `/obsidian-reindex` / `scripts/eval/semantic_search.py`.

### What problem would this solve?

`semantic_search.py` splits a note into `_CHUNK_CHARS`-sized pieces and embeds at most `_MAX_CHUNKS` of them, hardcoded at 8. Together those two constants set a fixed character budget per note, and text past it is not embedded.

The bound is deliberate and the default looks right. The comment above it says it "bounds time on huge notes," and that is a real cost worth bounding — a vault of atomic notes never approaches the limit, so the default costs those users nothing and protects their build time.

The issue is that the appropriate value is a property of the corpus, not of the tool. A vault of short atomic notes and a vault of long-form documents — research dossiers, book-length reference notes, meeting or interview transcripts — have genuinely different right answers, and the gap between them is large. For the second kind, a meaningful share of every long note is outside the budget and therefore unreachable by semantic search, no matter how good the ranking above it is. There is currently no way to move the value short of editing the file, which then fights every upstream merge.

### Proposed solution

Read the constant from the environment, default unchanged:

```python
_MAX_CHUNKS = int(os.environ.get("OBSIDIAN_EMBED_MAX_CHUNKS") or "8")
```

Default stays 8. The read follows the codebase's existing idiom for numeric settings - `int(os.environ.get(...) or "8")`, as in `OBSIDIAN_RRF_LEX_DEPTH` and `OBSIDIAN_SEARCH_MAX_FILES` - which also means an empty value falls back rather than raising. The name joins the `OBSIDIAN_EMBED_{BACKEND,URL,MODEL,KEY,EXCLUDE}` family, and it is documented with the other semantic-search knobs in the README.

**Deliberately not proposed:** raising the default. The build-time tradeoff belongs to the project and to the median vault. A vault that needs a different value can say so; everyone else is unaffected.

### Alternatives you considered

- **Split the long notes.** Sometimes right, but not always: a living dossier is often deliberately one consultable document, and reshaping content to fit an indexer's constant is the wrong direction.
- **`OBSIDIAN_EMBED_EXCLUDE`.** Removes whole notes from the index — the opposite of what is wanted.
- **Edit the constant locally.** Works, and is what prompted this: it is a permanent local diff that has to be re-applied after every upstream merge, which is exactly the kind of thing worth upstreaming instead.

### How does this interact with the AI-first vault rule?
Not applicable — no command or vault write is added or changed; this only affects how much of a note is embedded at index time.

### Additional context
Happy to open the PR: about five lines in `semantic_search.py`, plus a README knob line, a CHANGELOG entry, and two tests in `tests/test_index_robustness.py`.

etcook:
## What this PR does

Reads the per-note embedding cap from `OBSIDIAN_EMBED_MAX_CHUNKS` instead of hardcoding it at 8, leaving the default and every existing behaviour unchanged. `_CHUNK_CHARS` × `_MAX_CHUNKS` is a per-note character budget whose right value depends on how long the vault's notes are; long-form vaults can now set it without carrying a local patch, and nobody else sees a difference.

## Type of change

## Related issues

Closes #263

## How I tested it

- `tests/test_index_robustness.py` — two new cases: the default is pinned at 8 (the bound is deliberate, so the default must not drift), and the override is honoured for `64` and `1`, and an empty value falls back to the default.
- Full suite green locally: 777 passed, 1 skipped.
- Verified on a long-form vault that a raised cap embeds the full text of notes that were previously truncated, and that the default path is byte-identical — same constant, same value, read one level later.

## AI-first rule compliance

Not applicable — no vault-writing command is added or changed.


eugeniughelbur (commented):
## First review (COMMENT only — no approve, no merge)

**What is good**
- Small, focused change: `_MAX_CHUNKS` reads `OBSIDIAN_EMBED_MAX_CHUNKS` with default `8`, same `int(os.environ.get(...) or "N")` pattern as other numeric knobs.
- Default and behaviour stay unchanged for everyone who does not set the env.
- Tests pin the default at 8, honour overrides (`64`, `1`), and empty-string fallback — that matches the issue contract.
- README knob line + CHANGELOG entry present; closes #263 cleanly.

**What is risky / watch**
- CI status is still pending/empty on this head (`44a10b8`) — wait for smoke before merge yes.
- Non-numeric env values will raise on `int(...)` (same as sibling knobs; not introduced here).
- `importlib.reload` in the new tests is cleaned in `finally`; fine as long as the suite keeps that restore path.

**Verdict:** Looks merge-ready once smoke is green. Waiting on Eugeniu's yes — will not approve or merge from this triage.

PR #276, opened 2026-09-14
Health: keep the scan running when a wikilink is too long to be a filename

### LINKED ISSUE #272: [Bug] /obsidian-health crashes when a wikilink is longer than the filesystem allows for a name
### Which command broke?
/obsidian-health

### What happened?

One note containing a very long `[[...]]` stops the health scan for the entire vault. A captured web page with inline script left in it can produce one: something like `window.IJ_values = [[null,null,null,...]]` matches the wikilink pattern.

The link does not resolve to a note, so `check_wanted_notes` checks whether it names a folder with `(vault / link).is_dir()` (`scripts/vault_health.py:1005`). On Python 3.10 through 3.13, `pathlib.Path.is_dir()` raises `OSError` for a name over the filesystem limit instead of returning `False`, and nothing catches it. Python 3.14 changed `is_dir()` to return `False` on errors, so the crash does not reproduce there, but `requires-python` is `>=3.10` and a `uv` environment on 3.13 reproduces it.

The link reported as a wanted note, or skipped, and the rest of the scan completing.

### Steps to reproduce

1. Create an empty folder with one note, `scraped.md`:
   ```markdown
   ---
   type: source
   ---
   # Scraped page

   window.IJ_values = [[null,null,null, ... repeated until the text inside the brackets passes 255 characters ... ]];
   ```
2. Run `uv run python scripts/vault_health.py --path <folder>`.
3. The scan exits with the traceback below.

### Skill version
`fd19b7c`

### Operating system
macOS (APFS, 255-byte name limit). The same limit applies on ext4.

```shell
Traceback (most recent call last):
  File "scripts/vault_health.py", line 1299, in <module>
  File "scripts/vault_health.py", line 1279, in main
  File "scripts/vault_health.py", line 1185, in run_health_check
  File "scripts/vault_health.py", line 1005, in check_wanted_notes
  File ".../python3.13/pathlib/_abc.py", line 465, in is_dir
  File ".../python3.13/pathlib/_local.py", line 515, in stat
OSError: [Errno 63] File name too long: '<vault>/null,null,null,...'
```

### Suggested fix

Replace `(vault / link).is_dir()` with `os.path.isdir(vault / link)`, which returns `False` for any `OSError` on every supported Python version (checked on 3.10 and 3.13). Ordinary links behave as they do now, and the long one is reported as a wanted note. A regression case in `tests/test_vault_health_precision.py` with a 300-character link would pin it. Until then, `exclude-link-scan` on the offending file lets the scan finish.

etcook:
## What this PR does

`check_wanted_notes` calls `(vault / link).is_dir()` on every unresolved link in case it names a folder. On Python 3.10-3.13 that raises `OSError` once the name passes the filesystem limit, so one captured page with inline script in it ended `/obsidian-health` for the whole vault. The `OSError` is now treated as "not a folder", which is what Python 3.14 already does, and the link is reported as a wanted note.

The issue suggested `os.path.isdir`; the repo's ruff config flags that (PTH112, prefer pathlib), so this uses `try`/`except OSError` around `Path.is_dir()` instead, with the same behaviour.

## Type of change

## Related issues

Closes #272

## How I tested it

- New test `test_overlong_wikilink_is_a_wanted_note_not_a_crash` in `tests/test_vault_health_precision.py`: a note containing `[[null,null,...]]` well past 255 characters. It fails on current `main` with `OSError: [Errno 63] File name too long` and passes with the change, on both Python 3.11 and 3.13.
- Full suite: 838 passed, 1 skipped, 1 failed. The one failure, `test_research_cjk_and_models.py::test_default_walks_ladder_past_cohort_404`, fails the same way on unmodified `main` on this machine and is unrelated to this change.
- `ruff check .` clean.

## AI-first rule compliance

Not applicable: no vault-writing command is added or changed.

## Checklist


eugeniughelbur (commented):
First review (triage) — focused pass on the remote diff, no checkout.

**What is good**
- Matches #272 exactly: `Path.is_dir()` OSError on overlong names aborted the whole scan on Python 3.10–3.13; treating it as "not a folder" matches 3.14 and keeps pathlib (PTH112) instead of `os.path.isdir`.
- Minimal surface: one guarded call in `check_wanted_notes`, CHANGELOG under Unreleased, regression in `test_vault_health_precision.py` with a `[[null,null,...]]` past 255 chars.
- Smoke tests green; mergeable clean.

**What is risky / residual**
- Low. Only residual is that other `is_dir()` / `stat()` call sites on user-controlled path fragments could hit the same OSError class; out of scope for this PR, and this path was the known crash.
- Author notes one unrelated suite failure also present on `main` (`test_default_walks_ladder_past_cohort_404`); not introduced here.

**Verdict:** Looks good to merge from this pass. Holding for owner yes — will not merge until then.

PR #277, opened 2026-09-14
Health: measure semantic-index coverage over the notes the index should hold

### LINKED ISSUE #273: [Bug] /obsidian-health counts notes excluded by OBSIDIAN_EMBED_EXCLUDE as missing from the semantic index
### Which command broke?
/obsidian-health

### What happened?

`check_semantic_index` (`scripts/vault_health.py:805`) compares every note in the vault against the keys in `.obsidian-semantic-index.json`. It does not read `OBSIDIAN_EMBED_EXCLUDE`, so notes the index build skipped on purpose are counted as missing. Once an excluded folder passes `INDEX_STALE_PCT` (5% of the vault), the warning appears on every run, lists the excluded notes, and tells the user to rebuild the index, which cannot clear it because the rebuild excludes the same notes again. The only way to silence the warning is to stop excluding the folder.

Coverage measured against the notes the index is meant to hold, so a vault whose index is complete apart from its excluded paths reports no warning.

### Steps to reproduce

1. Create a vault with nine notes at the root and one at `Private/secret.md`.
2. Build the index with `OBSIDIAN_EMBED_EXCLUDE="Private/"`. (Without Ollama running, a hand-written `.obsidian-semantic-index.json` of the form `{"model": "bge-m3", "format": 2, "notes": {"note1.md": {"hash": "x", "title": "note1", "vecs": [[0.1, 0.2]]}, ...}}` holding the nine root notes produces the same result.)
3. Run `OBSIDIAN_EMBED_EXCLUDE="Private/" uv run python scripts/vault_health.py --path <vault> --json`.
4. The output contains:
   ```json
   {"type": "semantic_index", "severity": "warning", "message": "Semantic index covers 9 of 10 notes; 1 (10%) are missing and can only be found by literal word match. Rebuild: uv run python scripts/eval/semantic_search.py --path \"<vault>\" --build", "files": ["Private/secret.md"]}
   ```
   In the same environment, `semantic_search._excluded("Private/secret.md")` returns `True`.

### Skill version
`fd19b7c`

### Operating system
macOS

### Suggested fix

Drop excluded paths from `notes` before `check_semantic_index` computes coverage, using the same rule as `semantic_search._excluded` (`rel == p or rel.startswith(p)` over the comma-separated `OBSIDIAN_EMBED_EXCLUDE`). Reusing that function, or moving the prefix list into `vault_scan.py` where both scripts can import it, keeps the two from drifting apart. A case in `tests/test_index_staleness.py` alongside the existing coverage tests would cover it.

etcook:
## What this PR does

`check_semantic_index` counted notes excluded by `OBSIDIAN_EMBED_EXCLUDE` as missing from the index, so a vault whose excluded notes made up 5% or more of it got a warning on every run that the recommended rebuild could never clear. Coverage is now measured over the notes the build is meant to embed.

The prefix list and its match rule move into `scripts/vault_scan.py` (`embed_exclude_prefixes`, `is_embed_excluded`), and both `semantic_search.py` and `vault_health.py` use them, so the build and the check cannot disagree about which notes those are. `semantic_search.EXCLUDE_PREFIXES` and `_excluded` keep their names and behaviour.

## Type of change

## Related issues

Closes #273

## How I tested it

- Two new tests in `tests/test_index_staleness.py`, both failing on current `main` and passing with the change:
  - excluded notes are not reported, and the same vault without the exclude still warns;
  - with an exclude set, a real gap among the included notes still warns, the count excludes the carved-out folder ("10 of 20"), and the excluded note is not listed.
- The existing carve-out test (`test_smoke.py::test_semantic_search_math_and_carveout`, which reloads `semantic_search` under a new env value) and `test_index_robustness.py`, `test_exclude_policy.py`, `test_vector_math.py` pass.
- Full suite: 839 passed, 1 skipped, 1 failed, with the same unrelated failure as on `main`.
- `ruff check .` clean.

## AI-first rule compliance

Not applicable: no vault-writing command is added or changed.

## Checklist


eugeniughelbur (commented):
First review (triage) — focused pass on the remote diff, no checkout.

**What is good**
- Matches #273: coverage was comparing every note to the index, so `OBSIDIAN_EMBED_EXCLUDE` notes looked permanently missing and the suggested rebuild could never clear them.
- Shared `embed_exclude_prefixes` / `is_embed_excluded` in `vault_scan.py` used by both `semantic_search.py` and `vault_health.py` — build and check cannot drift. Wrappers `EXCLUDE_PREFIXES` / `_excluded` keep existing call sites.
- Denominator is `expected` (included notes only); empty-expected path short-circuits via `if not missing`. Tests cover "exclude silences false warn" and "real gap among included still warns (10 of 20) without listing Private/".
- Smoke tests green; mergeable clean.

**What is risky / residual**
- Low. Prefix match remains `rel == p or rel.startswith(p)` (same as before); users still need a trailing `/` on folder prefixes for the intended carve-out — pre-existing contract, not introduced here.
- Same unrelated `main` suite failure noted by author; not from this change.

**Verdict:** Looks good to merge from this pass. Holding for owner yes — will not merge until then.

PR #282, opened 2026-09-15
Let a Windows skill install register a SessionStart hook that runs (#281)

### LINKED ISSUE #281: [Bug] Windows: install.sh exits at the `python3` Store alias, and the SessionStart command it registers cannot run in Git Bash
### Which command broke?

install.sh

### What happened?

Found while verifying 0.16.0 on Windows for #269. #280 moved both hooks to an interpreter found by running it, but a skill install through `install.sh` still ends up with no working SessionStart hook on Windows, for two reasons.

1. **`install.sh` stops at `python3`.** Lines 74-75 still run `python3 "$SKILL_DIR/scripts/setup_settings_hook.py"` behind `command -v python3`. On a stock Windows install that name is the Microsoft Store alias, which passes the guard and exits non-zero (49 in Git Bash). `set -e` on line 3 then ends the installer right after "Registering session context hook...". No hook is registered, the research toolkit step and the closing instructions never run, and nothing says why.

2. **The command `setup_settings_hook.py` registers cannot run.** Line 25 is `HOOK_CMD = str(HOOK_PATH)`, which on Windows is `C:\Users\...\hooks\load_vault_context.sh` with no quotes. Claude Code runs hook commands through Git Bash there, and Git Bash reads every backslash as an escape. The same existing script fails in that form and runs with forward slashes (logs below). A headless Claude Code session given a SessionStart command in the backslash form recorded a `hook_non_blocking_error` with "No such file or directory" and never ran the script. So fixing the first problem alone would register a hook that fails at every session start.

The README FAQ says `install.sh` supports Windows (MSYS2/Git Bash).

- `install.sh` finishes on Windows and registers the SessionStart hook through a Python that actually runs. `scripts/python-interpreter.sh` already provides `osb_python`.
- With no working interpreter, it finishes and prints the hook to add by hand, as its fallback already promises, instead of exiting.
- The registered command is one Git Bash can run: forward slashes, quoted when needed.

### Steps to reproduce

1. Windows 10/11 with Python from python.org and Git Bash, where `where python3` shows `...\WindowsApps\python3.exe`, the Store alias.
2. In a checkout of v0.16.0, run the installer against a throwaway home so nothing real is touched:
   `mkdir -p /tmp/osb-home && USERPROFILE="$(cygpath -w /tmp/osb-home)" HOME=/tmp/osb-home bash install.sh`
   It ends after "Registering session context hook..." with exit 49, and `/tmp/osb-home/.claude/settings.json` is not created.
3. Print the command `setup_settings_hook.py` would register:
   `python -c "import sys; sys.path.insert(0, 'scripts'); import setup_settings_hook as s; print(s.HOOK_CMD)"`
4. Run an existing hook script in that backslash form through Git Bash, then with forward slashes:
   `bash -c "$(cygpath -w "$PWD/hooks/load_vault_context.sh")"` fails with "command not found", exit 127.
   `bash -c "$(cygpath -m "$PWD/hooks/load_vault_context.sh")" < /dev/null` prints the hook's JSON, exit 0.

### Skill version (release tag or commit SHA)

v0.16.0 (02fba47)

### Operating system

Windows

```shell
$ USERPROFILE="$(cygpath -w <home>)" HOME=<home> bash install.sh
Installing obsidian-second-brain...
...
Symlink failed (requires Developer Mode). For the cleanest setup,
clone the repo directly into the skills folder:
  git clone https://github.com/eugeniughelbur/obsidian-second-brain "<home>/.claude/skills/obsidian-second-brain"
Then re-run install.sh from that location.
Registering session context hook...
$ echo $?
49
$ test -f <home>/.claude/settings.json || echo "not written"
not written

$ python -c "import sys; sys.path.insert(0, 'scripts'); import setup_settings_hook as s; print(s.HOOK_CMD)"
C:\Users\<user>\...\.claude\skills\obsidian-second-brain\hooks\load_vault_context.sh

$ bash -c 'D:\<clone>\hooks\load_vault_context.sh'
bash: D:<clone>hooksload_vault_context.sh: command not found
$ echo $?
127
$ bash -c 'D:/<clone>/hooks/load_vault_context.sh' < /dev/null
{"hookSpecificOutput": {"hookEventName": "SessionStart", "additionalContext": "**Skill root** (obsidian-second-brain): ...
$ echo $?
0

Environment: Windows 10 Pro 10.0.19045, Git Bash, Claude Code 2.1.270, Python 3.10.4 (python.org), uv 0.12.6.
```

### Confirmation

i-so-late:
## What this PR does

Fixes the two problems in #281, which together leave a skill install on Windows without a working SessionStart hook.

## Type of change

- [ ] 📜 New slash command
- [ ] 📚 Documentation update
- [ ] 🧹 Refactor / cleanup

## Related issues

Closes #281 

## How I tested it

On Windows 10 Pro 19045 with Git Bash and Python 3.10 from python.org, where `python3` is the Store alias:

- **The installer, 0.16.0 against this branch,** each run against a throwaway `USERPROFILE`. 0.16.0 exits 49 right after "Registering session context hook..." and writes no `settings.json`. This branch exits 0, prints "SessionStart context hook added" and "Done.", and registers `C:/Users/.../hooks/load_vault_context.sh`.
- **The command form in a real session:** a headless session (`claude -p --settings ...`) was given two SessionStart commands, one in the form `str()` wrote and one with forward slashes. Only the forward-slash one ran; the transcript records a `hook_non_blocking_error` with "No such file or directory" for the other.
- **New tests, each failing before the change and passing after:**
  - `test_install_sh_resolves_the_interpreter_instead_of_running_python3`
  - `test_install_sh_registers_the_hook_when_python3_is_the_store_alias` runs the real installer against a throwaway home with `python3` shadowed by a stub that exits 9009, and asserts it finishes and registers the hook without a backslash.
  - `test_install_sh_finishes_and_names_the_hook_when_no_python_runs` shadows every candidate.
  - `test_the_registered_command_runs_in_the_shell_that_executes_it` runs `hook_command()` through bash from a path with a space in it.
- **Suite:** the full suite on Windows, run on this branch together with two other hook fixes, gives 870 passed, 3 skipped and 8 failed. Seven of the eight fail the same way on this machine without these changes: four symlink tests need a privilege Windows withholds, plus `test_an_untouched_note_is_rewritten` and the two post-write command tests. The eighth is `test_a_name_that_exists_but_does_not_run_is_passed_over`, which already fails on Windows on 0.16.0 because it looks for the Windows path of its stub in Git Bash output. `conformance_report.py --check` and `build_site.py --check` pass.
- **Lint:** `ruff check .` and `scripts/sweep_non_ascii.py --check` are clean.

- [ ] Frontmatter includes `ai-first: true`, `type:`, `date:`, `tags:`

## Checklist

- [ ] I updated the README or docs if user-facing behavior changed

## Screenshots / output (if visual or producing notes)

---


eugeniughelbur (commented):
First-pass review (COMMENT only — no merge from me).

**Good**
- Hits both #281 problems: installer uses `osb_python` instead of bare `python3`, and `hook_command()` uses forward slashes + `shlex.quote`.
- Real Windows smoke story in the PR body (throwaway USERPROFILE, Store alias, headless SessionStart).
- Solid regression tests: Store-alias PATH, no-Python fallback that still finishes, and bash run of a path with a space.
- CHANGELOG under Unreleased is clear.

**Risky / watch**
- `install.sh` still does unquoted `$PYTHON ...` (needed for `py -3` / `uv run`), so PATH quirks remain possible — same pattern as elsewhere, just note it.
- Author reports 8 local Windows failures (symlink privilege + known suite noise); none look introduced by this diff, but CI check runs are empty on this head so far.

**Checks:** no check runs / statuses on the head SHA yet (pending / empty).

**Merge:** looks ready from a first pass if Eugeniu is happy with the Windows verification. Owner: merge yes/no in chat.

i-so-late:
Thanks! I checked the CI status as well. The workflow is currently awaiting approval from the maintainer.

On the unquoted `$PYTHON`: `osb_python` only ever returns one of four fixed command names (`python3`, `python`, `py -3`, `uv run --no-project python`) and never a path, so word splitting only separates those words, and a PATH directory with spaces like `C:\Program Files` is never split because it is never part of the value.

eugeniughelbur:
Thanks — that clears the unquoted `$PYTHON` watch. `osb_python` returning only those four fixed forms means word-splitting is intentional and safe.

Check runs are still empty on this head from my side; approving the fork workflow so CI can run is the next owner step. Will wait for that (and Eugeniu's merge yes/no) before anything else.

PR #286, opened 2026-09-15
Schemas for company and tool entities, and a coverage fence that sees entity kinds named in prose

etcook:
## What this PR does

Implements the scope decided in #274. It includes everything in #284 (the company and tool schemas, their folder rows, and per-kind folders in `/obsidian-ingest`) and adds the parts #274 also asked for: the person schema's link, `vault-schema.md` reconciled to `type`, the sample vault fix, and the coverage test.

- `references/ai-first-rules.md` gains `type: company` and `type: tool`, in the shape of `type: person`; the tool schema has an optional `repo:` field, named like the project schema's. The person schema's `company:` becomes a bare `"[[Acme Corp]]"`, which resolves wherever the company note sits.
- The Entity Note in `references/vault-schema.md` discriminates on `type` instead of tags, and its "All entities" Dataview example filters on `type` to match.
- `references/folder-map.md` gives each entity kind its own row. Wiki-style keeps all three in `wiki/entities/`; Obsidian-style uses `People/`, `Companies/` and `Tools/`, and a note separates `Companies/` (companies the vault tracks) from `Businesses/` (companies the owner owns). The Obsidian-style column of the folder mapping table in `references/vault-schema.md` matches; `Jobs/` and `Businesses/` come off that row because they hold employment roles and owned companies, not tracked entities.
- `/obsidian-ingest` and `/obsidian-reconcile` resolve the entities folder per kind, since the new rows alone would not change where ingest files a company.
- `examples/sample-vault`: both person notes link `[[Currentscale Labs]]`, and the company now exists as a `type: company` example at `wiki/entities/Currentscale Labs.md`, so the four wanted-note findings are gone. The sample `_CLAUDE.md` folder list and `examples/README.md` are updated to match. The example sits in `wiki/entities/` because the sample vault already has `wiki/`, and wiki-style keeps every entity in that one folder.
- `tests/test_schema_coverage.py` gains `test_every_entity_kind_has_a_schema`. It reads the entity kinds from the `(entity)` rows of `references/folder-map.md`, and from any command that lists entity kinds as `for each a/b/c`, and requires a `type: x` schema heading in `references/ai-first-rules.md` for each. On current `main` it fails for `company` and `tool` from both sources.

Assumption: #274 settled on a company row, and I've given tools a `Tools/` row on the same pattern so a tool isn't filed in `People/` either. If you'd rather tools stay in `People/`, I'll take that row out.

[...]

PR #287, opened 2026-09-15
Setup script: build both vault layouts with --style wiki|obsidian

### LINKED ISSUE #283: [Bug] The documented default vault layout (wiki-style) cannot be created by any setup path
### Which command broke?
Other / not sure (`scripts/bootstrap_vault.py` and `/obsidian-init`)

### What happened?

The README's Vault Architecture section documents wiki-style as the default (`### Wiki-style (default) -- LLM-first`, `README.md:577`), and `references/folder-map.md` falls back to wiki-style folders when `_CLAUDE.md` is silent. No setup path creates that layout:

- `scripts/bootstrap_vault.py` creates Obsidian-style folders for every preset. None of them creates `wiki/`, the script takes no style argument, and its one call to `write_bases` passes `style="obsidian"` (line 1262). `WIKI_FOLDERS` (line 66) and the `style` parameter of `write_bases` (lines 173-179) are already there, but nothing reaches them.
- `/obsidian-init` creates no folders. It treats a vault as wiki-style only when a `wiki/` folder already exists at the root (`commands/obsidian-init.md:39`).

A wiki-style vault therefore has to be laid out by hand, and someone who follows the README's "No vault yet?" steps ends up with the layout the Vault Architecture section does not describe.

This came up before in #15, which found the README advertising a `--style` flag that was never implemented. It was closed by removing the example rather than adding the flag; the wiki-style section and the wiki-style defaults in `folder-map.md` stayed.

A smaller instance of the same drift: the builder preset's generated `_CLAUDE.md` says "New people mentioned → People/ (create stub if needed)", but that preset creates no `People/` folder and its folder map has no row for one. The creator preset has no `People/` either.

Either the setup script can create the layout the docs call the default, or the docs describe the layout the setup script creates.

### Steps to reproduce

1. `uv run python scripts/bootstrap_vault.py --path /tmp/v-default --name "Test User"`
2. `uv run python scripts/bootstrap_vault.py --path /tmp/v-builder --name "Test User" --preset builder`
3. Neither vault has a `wiki/` folder. `/tmp/v-builder` has no `People/`, while its `_CLAUDE.md` routes new people there.

### Skill version
`02fba47`

### Operating system
macOS

### Suggested fix

Two ways to close it:

- a) Add `--style wiki|obsidian` to `bootstrap_vault.py`, defaulting to `obsidian` so current output is unchanged. `wiki` creates the `wiki/` folders from the folder map and passes the style through to `write_bases`. This is option A from #15.
- b) Keep the script Obsidian-style only and bring the docs in line: the README says the script creates an Obsidian-style vault, wiki-style is described as a layout you set up yourself, and the wiki-style fallback in `folder-map.md` is reconsidered.

Separately from either, the builder and creator presets should either create `People/` or stop routing people to it. Happy to open the PR for whichever direction you prefer.

etcook:
## What this PR does

Implements the decision on #283 that Eugeniu and I discussed directly and I recorded in [a comment there](https://github.com/eugeniughelbur/obsidian-second-brain/issues/283#issuecomment-5680568528): keep both layouts and have `bootstrap_vault.py` build both.

- `--style wiki|obsidian`, defaulting to `obsidian`. With the default, every preset builds what it built before apart from one new line in `_CLAUDE.md` (see Implementation choices) and, for builder and creator, the people folder below.
- `--style wiki` resolves each preset folder to its wiki-style path from `references/folder-map.md` (`Daily` to `wiki/daily`, `People` to `wiki/entities`, `Ideas` and `Knowledge` to `wiki/concepts`, `Dev Logs` to `wiki/logs`, and so on), plus `boards/` and `templates/` from the wiki tree in `references/vault-schema.md`. The style is passed through to `write_bases`, so the Bases are stamped with the wiki paths.
- A preset folder the wiki layout does not rename keeps its own name under `wiki/`, lowercased and with each path segment slugged: `Sources/` to `wiki/sources/`, `Reading Queue/` to `wiki/reading-queue/`, `Finances/Spending/` to `wiki/finances/spending/`. `WIKI_PATHS` carries the genuine renames only, so a preset builds the same vault in either layout and a folder added to a preset later is not dropped from wiki-style vaults. Added in review: the first revision dropped unmapped folders, which made `--preset researcher --style wiki` build exactly what no preset at all would have built.
- The builder and creator presets now create `People/` (or `wiki/entities/`), since their own generated `_CLAUDE.md` sends new people there.
- `README.md`, `SKILL.md` and `references/vault-schema.md` document the flag and say which layout the script builds by default.

## Implementation choices

These go beyond what the decision spelled out, for you to confirm or change:

1. **Every path the script writes goes through one lookup, not only folder creation.** The generated `_CLAUDE.md` ("New people mentioned -> People/", board paths), `Home.md` links and its Dataview query, the template and seed-note queries (`FROM "Daily"`, `FROM "Tasks"`) and the board files all name folders. Creating `wiki/` folders alone would have produced a manual and dashboards pointing at folders the vault does not have, and a vault that fails its own health check.
2. **Two details of the fallback the review did not spell out.** Multi-word names are hyphenated (`Reading Queue/` to `wiki/reading-queue/`, not `wiki/reading queue/`), and an already-nested preset folder keeps its nesting (`Finances/Spending/` to `wiki/finances/spending/`, not `wiki/finances-spending/`). Both follow the existing wiki paths, which are lowercase single words with no spaces, but say the word if you want either flattened.
3. **Both styles record `- **Vault style:** obsidian|wiki` in `_CLAUDE.md`.** `folder-map.md` rule 3 decides where to create a missing folder from what `_CLAUDE.md` says about the layout, so stating it removes the guess. This is the one line added to every preset's default output.
4. **`Home.md` is still written in both styles**, even though the wiki tree in `vault-schema.md` does not list it, since it hosts the board links and the Bases-backed views.

Not in this PR: a style question in `scripts/quick-install.sh` (it still builds Obsidian-style), and creating `raw/`, `index.md` or `log.md`, which other commands create in both layouts.

While testing every preset in both layouts I found two defects that already exist on `main`, which I'll fix in a separate PR:

- The researcher preset seeds `Reading Queue/_Queue.md` with no incoming link, so a fresh Obsidian-style researcher vault fails `vault_health.py` with one orphan. The inspection test marks that one case `xfail(strict=True)` with the reason. A wiki-style vault no longer has it: its `Home.md` now links a preset's seed notes, as the Obsidian-style one always has.
- The executive, creator and researcher manuals route dev work to `Dev Logs/`, which those presets never create. The new test checks only the people route for that reason.

## Type of change

- [x] 📚 Documentation update

## Related issues

Closes #283

## How I tested it

- `tests/test_bootstrap_style.py` (new): no flag and `--style obsidian` build byte-identical vaults; every preset in wiki style creates each mapped folder and none of the Obsidian-style ones, records the style and stamps wiki paths into its Bases; no preset folder is dropped by the wiki layout, for every preset; the slug rule is pinned per folder name; every preset in both styles passes `vault_health.py` with zero findings (researcher/obsidian aside, above), a clean `freshness_lint.py`, and a folder map whose paths exist; builder and creator create their people folder in both styles; and `WIKI_PATHS` cannot drift from `references/folder-map.md`. The file cannot import against current `main`, which has no `STYLES` or `resolve_folder`.
- Default output compared against current `main` for all five presets and assistant mode: the only differences are the `Vault style` line, and for builder and creator the `People/` folder, its folder-map row, its `Home.md` link and `Bases/People.base`.
- Whole vault trees diffed for all five presets against the commit before the fallback: Obsidian-style output is byte-identical, so the fallback changed wiki-style vaults only.
- Full suite: 899 passed, 1 skipped, 1 xfailed (the researcher case above), 1 failed. The one failure, `test_research_cjk_and_models.py::test_default_walks_ladder_past_cohort_404`, fails the same way with this branch's own changes reverted and is unrelated.
- `ruff check .`, `sweep_non_ascii.py --check`, `build_site.py --check` and `conformance_report.py --check` pass.

## AI-first rule compliance

- [ ] If I added or changed a command that writes to the vault, the produced note has a `## For future agent` preamble

## Checklist


eugeniughelbur (commented):
First-pass review (COMMENT only — no merge from me).

**Good**
- Implements the #283 decision: `--style wiki|obsidian`, default `obsidian`, wiki paths from folder-map, Bases get the style, builder/creator create People (or wiki/entities).
- Docs no longer call wiki the default; README/SKILL/vault-schema updated.
- Strong test file `test_bootstrap_style.py` (default identity, wiki-only layout, health/freshness inspection, People for builder/creator, WIKI_PATHS vs folder-map).
- Smoke tests green.

**Needs owner confirm (from PR body — please answer these)**
1. Every written path goes through one lookup (manual, Home, templates, boards) — not only mkdir. Looks right to me; confirm.
2. **Topic folders skipped in wiki style** (Goals/, Health/, Finances/, Jobs/, Businesses/, Mentions/, etc. + their seed notes). Biggest product call — does this match intent?
3. Both styles write `- **Vault style:** …` in `_CLAUDE.md`.
4. `Home.md` still written in wiki style even though wiki tree in vault-schema does not list it.

Also noted out of scope: quick-install still Obsidian-only; pre-existing researcher orphan + Dev Logs route bugs deferred.

**Checks:** Smoke tests — success.

**Merge:** hold for owner answers on the four choices (esp. #2 topic folders). After that, looks mergeable. Owner: confirm choices + merge yes/no in chat.

eugeniughelbur:
Thanks for this. #283 is a real gap and the diagnosis is right: `write_bases` hardcoded `style="obsidian"` and every preset folder list was Obsidian-style, so the layout the docs led with could not be built. Dropping "default" from the wiki-style headings in the README and `references/vault-schema.md` is the honest fix for that half.

One thing to settle before I merge, and I think it is a product call rather than a defect.

**A preset stops meaning anything under `--style wiki`.** I ran the branch:

```
python3 scripts/bootstrap_vault.py --path /tmp/v --name "Test User" --preset researcher --style wiki
```

Folders created:

```
_trash  boards  templates
wiki/concepts  wiki/daily  wiki/entities  wiki/projects
```

The researcher preset declares `Daily, Sources, Literature, Hypotheses, Methodology, Synthesis, Reading Queue, Projects, People, Boards, Templates, _trash`. Five of those have no `WIKI_PATHS` entry and are dropped: `Sources/`, `Literature/`, `Hypotheses/`, `Methodology/`, `Reading Queue/`. What is left is what `--style wiki` with no preset at all would give you. Same shape elsewhere: `builder` loses `Architecture/` and `Debugging/`, `creator` loses all three `Content/` folders plus `Audience/` and `Publishing/`, and `default` loses ten of its twenty-one, including `Goals/`, `Jobs/`, `Businesses/` and `Finances/Spending/`. The seed notes that live in those folders go too, since `_writer` returns early when the top folder has no home.

Your reasoning holds as far as it goes. I grepped `commands/` and no command writes into any of them, so nothing breaks. But a preset is a promise about the shape of the vault, and someone who types `--preset researcher --style wiki` gets that promise dropped without being told. That is the shape of failure we just shipped a release about.

I would rather it did one of these, and I lean towards the first:

1. Fall back to `wiki/<lowercased name>` for a folder with no explicit mapping, so `Sources/` becomes `wiki/sources/` and the preset survives the layout switch. `WIKI_PATHS` then only carries the genuine renames, which is a shorter table and a clearer one.
2. Keep the drop, but print which folders were omitted and why, so it is a stated tradeoff and not a silent one. Document it in the `--style` help text too, not only in the PR body.

Option 1 also removes the question of what happens to the next preset folder someone adds, which under the current shape is dropped from wiki-style vaults by default and silently.

Everything else reads well: `resolve_folders` collapsing `Ideas` and `Knowledge` onto `wiki/concepts` is right, rewriting the Dataview `FROM` sources is the part that would have been easy to miss, and recording `Vault style:` in `_CLAUDE.md` gives `references/folder-map.md` something to resolve against. Adding `People/` to the builder and creator presets is a good catch on its own.

CI is green here. Once the folder question is settled I will merge this ahead of #286, since #286 needs `Companies/` and `Tools/` added to the presets and that is this file.

etcook:
Went with option 1. A preset folder with no `WIKI_PATHS` row now keeps its own name under `wiki/`, lowercased, each segment slugged: `Sources/` to `wiki/sources/`, `Reading Queue/` to `wiki/reading-queue/`, `Finances/Spending/` to `wiki/finances/spending/`. `WIKI_PATHS` is left holding the genuine renames only.

Your researcher example now builds `wiki/sources`, `wiki/literature`, `wiki/hypotheses`, `wiki/methodology` and `wiki/reading-queue` alongside the mapped four, and the next folder added to a preset is carried rather than dropped.

Three things followed from the folders existing, all in the commit:

- The seed notes that live in them are written again, so the wiki-style `Home.md` links a preset's seed notes the way the Obsidian-style one always has. Without that a fresh wiki vault would ship notes nothing links and fail its own health check. This also means wiki-style researcher no longer has the `_Queue.md` orphan; the Obsidian-style one still does, which is the pre-existing bug, still `xfail(strict=True)` and still headed for its own PR.
- The generated manual routes mentions and finances work to those folders when the preset carries them, instead of suppressing those routes for the whole layout.
- The `_CLAUDE.md` folder map describes a folder by the preset folder it came from, since `wiki/sources/` on its own says nothing about what belongs in it.

On rule 4: bootstrap creating `wiki/sources/` is a creation decision rather than a resolution one, so I added a line to `references/folder-map.md` saying so, and that the vault's `_CLAUDE.md` map is what commands resolve against afterwards. If you would rather the rule stay unqualified, I can move that note to the script instead.

Obsidian-style output is unchanged: I diffed whole vault trees for all five presets against the previous commit, byte-identical. New tests cover the no-drop promise per preset and the slug rule. `--style` help text and the README say what happens to an unmapped folder.

eugeniughelbur:
Option 1 landed well. Merging.

I checked the "Obsidian-style output is unchanged" claim rather than taking it, because you measured against your own previous commit and what matters to existing users is against `main`. Diffing all five preset trees against `main`:

- `default`, `executive`, `researcher`: one line, `- **Vault style:** obsidian`.
- `builder`, `creator`: that line, plus `People/`, `Bases/People.base` and the `Home.md` link.

Both are documented and intended, and nothing else moved. The claim holds.

The thing I asked for is there: `--preset researcher --style wiki` now builds `wiki/sources`, `wiki/literature`, `wiki/hypotheses`, `wiki/methodology` and `wiki/reading-queue` beside the mapped four, so the preset means something under either layout. A freshly bootstrapped wiki vault reports 0 health issues on `researcher`, `default` and `creator`, so the showroom rule holds for the new layout too. 945 tests pass.

Writing the seed notes back was the right call to notice. A layout that creates folders and then ships nothing in them would have failed its own health check, which is the failure this repo keeps naming.

On rule 4: keep the note in `references/folder-map.md` where you put it. Bootstrap is the one caller that creates rather than resolves, and the rule is easier to follow with the exception written next to it than with it implied. Leaving it unqualified would just mean the next reader has to rediscover why the script looks like it breaks the rule.

The `_Queue.md` orphan staying `xfail(strict=True)` and headed for its own PR is the right split.

On #286: your reading is correct, take it as a rule rather than the list as it stands on `main` today. With this PR adding `People/` to builder and creator that means all five presets, which is what I want. A rule tied to `People/` stays true the next time someone edits a preset; a list of three quietly goes stale. A builder vault tracking the tools it uses is also just reasonable.

Go ahead and rebase #286 onto this.

One more, unrelated to your PRs but adjacent to your layout work: #292 reports that `check_orphans` exempts dated folders through a hardcoded Obsidian-style list, so `wiki/daily/YYYY-MM-DD.md` rings every day in a wiki-style vault and `Logs/` rings in both. Same wiki-style drift class as #283. I am taking that one, so it will not collide with your rebase.

Issue #269, opened 2026-09-13
[Bug] Windows: hooks run `python3`, which is the Microsoft Store stub - SessionStart injects nothing and validate-ai-first.sh skips its Python checks, both silently


i-so-late:
### Which command broke?

Other / not sure

### What happened?

Not a slash command - the two hooks, on a normal Windows install.

`hooks/hooks.json` (line 9) runs `python3 "${CLAUDE_PLUGIN_ROOT}/hooks/load_vault_context.py"` at SessionStart. On Windows, `python3` resolves to the Microsoft Store "App Execution Alias" stub (`%LOCALAPPDATA%\Microsoft\WindowsApps\python3.exe`): it prints nothing and exits non-zero (9009 from cmd/PowerShell, 49 under Git Bash, which is where Claude Code runs hooks). So the hook injects zero characters - no **Skill root**, no `_CLAUDE.md`. The only trace is a `hook_non_blocking_error` attachment in the session transcript ("Failed with non-blocking status code: No stderr output"); nothing is shown in the UI.

`hooks/validate-ai-first.sh` guards checks 5, 6 and 7 (banned non-ASCII characters, secrets, tag syntax) with `command -v python3` (lines 254, 305, 346). The guard passes because the stub exists, `python3 - "$FILE" <<'PYEOF'` produces no output, and the three checks are skipped without a word. Checks 1-4 are pure bash and still fire, so the hook looks alive. Verified below: a note with an em-dash and a fake `sk-` key passes silently as shipped and produces both warnings once `python3` points at a real interpreter.

The same name is hardcoded in `scripts/setup_settings_hook.py:21` (`HOOK_CMD = f"python3 {HOOK_PATH}"`), `install.sh:74-75` and `scripts/setup.sh:148`.

Why I think this is the plugin's to fix: the python.org installers - the default way to get Python on Windows - ship `python.exe` and `py.exe`, never `python3.exe`; that name belongs to the Store alias out of the box. So on a typical Windows machine `python3` exists but does nothing, `command -v` cannot tell, and the Windows support that landed in v0.15 (#240-#248) still leaves the session-context hook and a third of the validator dead. The research scripts are unaffected because they go through `uv run`.

### What did you expect to happen?

- The SessionStart hook injects the Skill root and, inside the vault, `_CLAUDE.md` - on Windows as on macOS and Linux.
- validate-ai-first.sh runs all seven checks, or says which ones it could not run.
[...]

Issue #160, opened 2026-07-22
[Bug] The MCP server ignores OBSIDIAN_VAULT_PATH from ~/.config/obsidian-second-brain/.env - same root cause as #124, fixed for the research scripts but not the MCP path


SylvesterTee:
### Which command broke?

Other / not sure - the MCP server (`integrations/obsidian-mcp-server/`), so every `obsidian_*` tool.

### What happened?

Every `obsidian_*` MCP tool call fails with:

```
Error executing tool obsidian_search: OBSIDIAN_VAULT_PATH is not set
```

even though `OBSIDIAN_VAULT_PATH` **is** set, in the documented location:

```
~/.config/obsidian-second-brain/.env
```

`architecture.md` states:

> Configuration is read from the environment or `~/.config/obsidian-second-brain/.env`; `OBSIDIAN_VAULT_PATH` is required so notes save to the right vault on any machine.

The MCP server honours only the first half of that sentence.

The practical effect is a silent downgrade: the agent falls back to plain file reads/grep and keeps working, so nothing looks broken - the vault search simply never runs, and the ranking, snippets, freshness rerank and supersession logic in `vault_ops.search()` are all bypassed.

### Root cause

- `integrations/obsidian-mcp-server/vault_ops.py:23` resolves the vault from `os.environ` only. Grepping that whole directory returns **zero** references to `dotenv`.
- By contrast `scripts/research/lib/config.py:4,10` **does** call `load_dotenv(ENV_PATH)`.
- `.claude-plugin/plugin.json` declares the server with no `env` block:

```json
"mcpServers": {
  "obsidian-second-brain": {
    "command": "uv",
    "args": ["run", "--with", "mcp", "python", "${CLAUDE_PLUGIN_ROOT}/integrations/obsidian-mcp-server/server.py"]
  }
}
```

while `integrations/obsidian-mcp-server/README.md` shows an example config that *does* include `"env": { "OBSIDIAN_VAULT_PATH": "/path/to/your/vault" }`.

So a plugin-marketplace install whose config lives in `.env` gets a completely non-functional MCP server, with an error message that contradicts what the user actually configured and what the docs told them to do.

This is #124 (*"/research and /research-deep silently ignore PERPLEXITY_API_KEY from .env"*) in a different code path - that fix landed in the research toolkit, and the MCP server was not swept.

### What did you expect to happen?

[...]

PR #288, opened 2026-09-15
Fix SessionStart hook silently skipping the vault manual (env-only OBSIDIAN_VAULT_PATH check)

### LINKED ISSUE #285: load_vault_context.py SessionStart hook never injects _CLAUDE.md (OBSIDIAN_VAULT_PATH env-only check, same bug class as #160/#124)
## Bug

`hooks/load_vault_context.py`'s `vault_manual_block()` reads the vault path with:

```python
vault = os.environ.get("OBSIDIAN_VAULT_PATH", "")
if not vault:
    return ""
```

A marketplace/plugin install only writes `OBSIDIAN_VAULT_PATH` into `~/.config/obsidian-second-brain/.env` — it never exports it as a real process environment variable. So this check fails on every single session, including one whose cwd genuinely is the vault, and the SessionStart hook silently falls back to emitting only the "Skill root" block. `_CLAUDE.md` never gets injected at all, for anyone, ever.

This is the same root cause already fixed twice elsewhere in this codebase — `hooks/validate-ai-first.sh` has an explicit comment about it:

> "Environment wins; fall back to the documented config .env, because a plugin-marketplace install configures the vault there and never exports the variable - so an env-only check made this hook a silent no-op for exactly the installs that need it most. Same root cause as #160 (MCP server) and #124 (research toolkit); this is the third code path, swept when the hook turned out never to have been wired at all."

`load_vault_context.py` appears to be the missed fourth instance of this exact bug class.

## Impact

For a multi-project setup (one shared vault, several separate Claude Code sessions in different project directories all writing into it via the `obsidian_*` MCP tools), this means:
- The vault's own `_CLAUDE.md` operating rules (AI-first note rules, stub-on-dangling-link, etc.) never reach any session automatically, including a session working from inside the vault itself.
- Sessions write into the vault "blind," which in practice produces exactly the failure modes the AI-first rules exist to prevent: dangling wikilinks left unstubbed for days, and at least one note that shipped with a duplicated/corrupted frontmatter block from a template pasted over existing content.

## Suggested fix

Add the same `.env` fallback `validate-ai-first.sh` already uses, and (separately) inject a short pointer block instead of nothing for sessions whose cwd is outside the vault but that still have the vault configured (a single machine-wide `OBSIDIAN_VAULT_PATH` means sibling-project sessions have the tools available without ever being "in" the vault) — the full manual stays cwd-gated to avoid paying its token cost on every unrelated session on the machine.

```python
def resolve_vault_path() -> str:
    """OBSIDIAN_VAULT_PATH from the environment, falling back to the documented
    config .env. A marketplace/plugin install configures the vault there and
    never exports it as a real process env var, which made this hook's vault-
    manual injection a silent no-op on every session, including one whose cwd
    IS the vault - the env-only check never once passed. Same root cause the
    bash hook (validate-ai-first.sh) already had to patch for twice (#160, the
    MCP server; #124, the research toolkit); this Python hook was the missed
    third code path.
    """
    vault = os.environ.get("OBSIDIAN_VAULT_PATH", "")
    if vault:
        return vault
    env_file = os.environ.get("OBSIDIAN_ENV_FILE") or str(
        Path.home() / ".config" / "obsidian-second-brain" / ".env"
    )
    try:
        for line in Path(env_file).read_text(encoding="utf-8").splitlines():
            line = line.strip()
            if line.startswith("OBSIDIAN_VAULT_PATH="):
                vault = line.split("=", 1)[1].strip().strip('"').strip("'")
    except OSError:
        pass
    return vault
```

Then in `vault_manual_block()`:
- call `resolve_vault_path()` instead of the direct `os.environ.get(...)`
- when the path resolves but cwd doesn't match the vault, return a short pointer block (vault root path + "read `_CLAUDE.md` Section 0 before writing here") instead of `""`

I patched a local install this way and confirmed both code paths (cwd inside vault -> full manual; cwd elsewhere with the vault configured -> short pointer) work correctly, tested via direct stdin payloads to the hook script. Happy to open a PR with this exact change if useful — flagging as an issue first in case there's a preferred approach for the pointer-block piece.

(Found and reproduced while debugging why a shared vault's other project sessions never seemed to follow the AI-first note rules, using Claude Code.)

Arman-no:
## What

`load_vault_context.py`'s `vault_manual_path()` read `OBSIDIAN_VAULT_PATH` only via `os.environ.get(...)`. A marketplace/plugin install only writes it into `~/.config/obsidian-second-brain/.env` and never exports it as a real process env var, so the check failed on every session - including one whose cwd genuinely is the vault. The vault manual never actually loaded for anyone via this hook.

Same root cause as #124, #160, #269. Closes #285.

## Fix (updated after #289)

Rebased onto `main` now that #289 landed. `vault_manual_path()` now calls `osb_env.vault_path()` instead of reading `os.environ` directly - a three-line diff instead of the standalone fallback I originally wrote, per your comment on #285.

## Tests

Added two tests to `tests/test_session_context_cap.py`, since #289's own tests cover the resolver in isolation but nothing exercised the hook end to end against this exact failure:

- A vault configured only in the `.env` file (no real process env var) still gets its manual injected.
- A config file that does not exist yet leaves the hook silent, not raising.

Also added the `CHANGELOG.md` entry under "Unreleased".

## Verified

- `uv run pytest tests/test_session_context_cap.py tests/test_osb_env.py -v`: all 9 tests in the first file pass (including both new ones); 40 of 42 pass in the second. The 2 failures there are pre-existing on a clean `main` too (confirmed by stashing my change and re-running) - both are Git-Bash-on-Windows path-format quirks in the bash half's own tests, unrelated to this change.
- Running the full suite locally now for a final check before merge.

## Scope

Still just this fix. Left out the earlier idea of a short pointer block for sibling-project sessions outside the vault cwd - that's a separate design question you wanted to think through before it's baked in, happy to open it as its own follow-up if/when you want it.

---
Thanks for the thorough review, and for opening #289 off the back of it. Happy to keep helping out on this if useful - it's a project I actually use.


eugeniughelbur:
Thanks, the fix is right and the diagnosis in #285 was right. I checked the mechanism before commenting: `scripts/setup.sh` writes the vault path in two places, the `env` block of `~/.claude/settings.json` (line 96) and `~/.config/obsidian-second-brain/.env` (line 109). Only the first becomes a real process env var. An install that writes the second and not the first leaves `vault_manual_path()` returning `None` on every session, including one whose cwd is the vault. Your fallback closes that.

`Path.home()` is also the right call here and not the Windows bug it looks like: `scripts/platform-home.sh` documents that Python's `Path.home()` already resolves to `USERPROFILE` on Windows shells, which is why only the bash half needs `osb_platform_home`.

Three things before this goes in.

**1. It needs a test.** This is the fourth time this bug class has shipped: #124, #160, #269, and now #285. Every one of them was a hook that ran, found nothing, and said nothing. A test that writes a `.env` with `OBSIDIAN_VAULT_PATH`, leaves the variable unset in the environment, and asserts the hook injects the manual is what stops the fifth. `tests/test_python_interpreter.py` has the shape to copy: it shadows binaries and runs the real thing against a throwaway home. Please also cover the case the fallback must not break, an `OBSIDIAN_ENV_FILE` that does not exist, where the hook has to stay silent rather than raise.

**2. It needs a CHANGELOG entry** under "Unreleased", in the style of the surrounding ones.

**3. This is the seventh copy of the same six lines.** The `OBSIDIAN_ENV_FILE` or `~/.config/obsidian-second-brain/.env` resolution is already written out in `hooks/validate-ai-first.sh:207`, `scripts/setup.sh:27`, `scripts/research/lib/config.py:13`, `scripts/research/lib/source_config.py:33` and `scripts/eval/retrieval_eval.py:58`. `scripts/eval/behavior_eval.py:69` is a sixth that quietly ignores `OBSIDIAN_ENV_FILE` altogether, which is exactly how these copies drift.

The reason this bug keeps recurring is that there is no one place to fix it. I would rather this PR added the seventh copy knowingly and opened a follow-up issue to extract a shared resolver, than added it silently. If you would rather do the extraction here, I will take that instead, as long as `validate-ai-first.sh` keeps its inline copy for the reason `platform-home.sh` documents: it gets copied into other harnesses by hand and has to stand alone.

Happy to merge once 1 and 2 land. 3 can be a follow-up with your name on it if you want it.

eugeniughelbur:
Follow-up on point 3: I opened #289, which extracts the shared resolver rather than leaving it as a follow-up issue. Your diagnosis in #285 is what made the case for it, so thank you for that.

Once #289 lands, this PR gets smaller rather than larger. `resolve_vault_path()` becomes:

```python
sys.path.insert(0, str(Path(__file__).resolve().parents[1] / "scripts"))
import osb_env

# ... in vault_manual_path()
vault = osb_env.vault_path()
```

`osb_env.vault_path()` already does what your function does, plus a few things it did not: it ignores an empty `OBSIDIAN_VAULT_PATH` left behind in the environment rather than treating it as configured, it handles `export ` prefixes and CRLF, and it is stdlib-only so the hook keeps working on a machine with nothing installed.

What is still yours, and still the part that matters:

- The **test**. Nothing in #289 covers the hook, because #289 does not touch the hook. A test that writes a `.env`, leaves `OBSIDIAN_VAULT_PATH` unset in the environment, and asserts the manual is injected is what closes #285 for good. Please still add it.
- The **CHANGELOG** entry.

I did not touch `hooks/load_vault_context.py` on purpose, so this stays your fix. Rebase on `main` once #289 is merged and it should be a three-line diff plus the test.

Sorry for moving the ground under you mid-review. The seventh copy was the thing that convinced me it needed doing now rather than later, and finding it was a direct result of your issue.

Arman-no:
Updated: rebased onto main now that #289 landed, so this is the 3-line diff you predicted, plus the two tests and the CHANGELOG entry you asked for. Details in the PR description above.

Thanks for taking the time on the review, and for #289 - genuinely appreciate it. Feel free to credit this by name (Arman) rather than a handle if that's easier for you. This is a tool I actually use daily, so happy to keep contributing if you'd find that useful.

Arman-no:
Ran the full suite properly (in a separate worktree against clean \`main\`, so nothing carried over): 55 failed, 849 passed there. Same branch, my change: 55 failed, 851 passed - the +2 is exactly the two tests I added. Diffed both failure lists directly, byte-for-byte identical set of 55, so nothing here is a regression from this change.

PR #306, opened 2026-09-20
Scope the .obsidian skip to the vault's own config directory

jameswolensky:
## What this PR does

A vault kept under a folder named `.obsidian` had every note skipped by the write-time hook, silently. This scopes the skip entry to the vault's own config directory.

## Type of change

- [ ] 📜 New slash command
- [ ] 📚 Documentation update
- [ ] 🧹 Refactor / cleanup

## Related issues

n/a

## What I found

My vault is at `~/.obsidian/life-os`. I noticed the hook had never warned me about anything and went looking.

- `hooks/validate-ai-first.sh` skips a set of non-first-class paths. One entry is `*/.obsidian/*`, meant for the vault's config directory.
- The pattern matches that segment anywhere in a path, not under the vault root. Every note in my vault sits below a folder called `.obsidian`, so every note matched and the hook returned before any check ran.
- Nothing errored and nothing printed. The hook exits 0 whether it validated a note or skipped it, so the vault looked exactly like one that was passing. No frontmatter check, no preamble check, no banned-character check, for as long as the vault has existed.

Obsidian allows a vault inside `.obsidian`, and a vault named after the app's config folder is an easy layout to land on.

## The change

```diff
-  */raw/*|*/templates/*|*/_export/*|*/.obsidian/*|*/.git/*|...
+  */raw/*|*/templates/*|*/_export/*|"$VAULT_KEY"/.obsidian/*|*/.git/*|...
```

`$VAULT_KEY` rather than `$VAULT`, because `path_key()` lowercases on Windows and the block sets `nocasematch` there. The unfolded path would stop matching the folded `$FILE_KEY` on the one platform the surrounding code already takes care over.

The other entries are left alone. They are relative segments like `*/raw/*` and carry no vault-root ambiguity.

## How I tested it

`tests/test_validate_hook_vault_scope.py`, five tests:

- A note in a vault under `.obsidian` is checked.
- The vault's own config directory is skipped, in both a plain vault and a nested one.
- An ordinary note is still checked.
- A `.obsidian` folder deeper in the vault is a user folder and gets checked.

Two of the five fail against the old pattern, so they pin the bug rather than the fix.

Full suite on macOS, Python 3.14: 976 passed, 1 skipped, 1 xfailed.

I have not tested this on Windows. The `$VAULT_KEY` choice is read from `path_key()` and the `nocasematch` block, not from a run there.

- [ ] Frontmatter includes `ai-first: true`, `type:`, `date:`, `tags:`

No command changed and nothing here writes to the vault.

## Checklist

- [ ] I updated the README or docs if user-facing behavior changed

No user-facing behavior changed, so the README is untouched.


eugeniughelbur (commented):
First review (comment only — not a merge).

**What looks solid**
- Real silent failure: vault under a path segment named `.obsidian` (e.g. `~/.obsidian/life-os`) matched bare `*/.obsidian/*`, so every note was skipped and the hook still exited 0.
- Scoping the skip to `"$VAULT_KEY"/.obsidian/*` matches how the surrounding block already folds paths on Windows (`path_key` + `nocasematch`).
- The five tests in `tests/test_validate_hook_vault_scope.py` pin the bug (two fail on the old pattern) and keep the vault's own config directory skipped while a deeper user `.obsidian/` folder still gets checked.
- CHANGELOG Fixed entry is present. Smoke tests on this head were green.

**Before merge**
- Branch is now **dirty against `main`** after [#305](https://github.com/eugeniughelbur/obsidian-second-brain/pull/305) landed (docs-command fence). Please rebase/update the branch and re-run CI.
- I have not re-run the suite on Windows; the `$VAULT_KEY` choice is read from existing path-folding code, as you noted.

Holding for owner merge yes after a clean rebase.

